PR #2, opened 2021-03-10
Api get set opts

qzhuyan:


z8674558 (approved):
getopt covers sockname?

qzhuyan:
> getopt covers sockname?

yes, but for now, I will leave the getopt/2 to return raw binary. We can decide later if we need to merge getname and getopt.

PR #5, opened 2021-03-12
Bring latest msquic

z8674558:
v1.0.0 does not support MacOS.

### DIFF
--- a/c_src/quicer_config.c
+++ b/c_src/quicer_config.c
@@ -61,7 +61,7 @@ NewCredConfig(ErlNifEnv *env, const ERL_NIF_TERM *option)
       return NULL;
     }
 
-  Config = (QUIC_CREDENTIAL_CONFIG_HELPER *)QUIC_ALLOC_NONPAGED(
+  Config = (QUIC_CREDENTIAL_CONFIG_HELPER *)CXPLAT_ALLOC_NONPAGED(
       sizeof(QUIC_CREDENTIAL_CONFIG_HELPER), QUICER_CREDENTIAL_CONFIG_HELPER);
 
   memset(Config, 0, sizeof(QUIC_CREDENTIAL_CONFIG_HELPER));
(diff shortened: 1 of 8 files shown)


test/quicer_SUITE.erl
```diff
@@ -188,7 +188,6 @@ tc_stream_client_init(Config) ->
     listener_ready ->
       {ok, Conn} = quicer:connect("localhost", Port, [], 5000),
       {ok, Stm} = quicer:start_stream(Conn, []),
-      {ok, {_, _}} = quicer:sockname(Stm),
```
qzhuyan, line 191: why removed?

qzhuyan (dismissed):
when merging, pls specify the msquic version as v1.1.0.

test/quicer_SUITE.erl
z8674558: Previouosly quicer:sockname(Stream) returned sockname of Connection. But we can assume we always have Connection which has created Stream?

qzhuyan: Yes. The stream should always have a connection that runs on.

z8674558: I thought we should only allow sockname(Connection) or sockname(Listener), because sockname(Stream) was actually sockname(Connection).

qzhuyan: I agree in terms of calling msquic API.

But in reality,  The application build on top of quicer might assume the same behavior as when they deal with socket, 
like the following:

```
inet:sockname(socket())
```

So they expect to call:

```
quicer:sockname(stream_handler())
```

I think it is ok to return the sock name of connection even the argument is a stream. 

In the stream ctx, there is a pointer to the connection ctx that can be referred to.

z8674558: Now I got it. fixed

qzhuyan (approved):

PR #7, opened 2021-03-13
chroe(format): Arguments Packing

qzhuyan:
Add two overrides in dot clang-format file.

BinPackArguments: false
BinPackParameters: false

This PR is a result of running clang-format in-place updates. 
We need somehow define and automate our code style.


z8674558 (approved):
maybe clang-format in github actions?

qzhuyan:
> maybe clang-format in github actions?

Good point, I will add clang-format in github action for this branch only.

PR #8, opened 2021-03-15
feat(api): quicer:getstat/2 part 2

qzhuyan:
Support getstat on stream but fallback to getstat on connection.

### DIFF
--- a/c_src/quicer_config.c
+++ b/c_src/quicer_config.c
@@ -335,6 +335,13 @@ getopt3(ErlNifEnv *env, __unused_parm__ int argc,
     }
   else if (IS_SAME_TERM(eopt, ATOM_QUIC_PARAM_CONN_STATISTICS))
     {
+      if (q_ctx && Level == QUIC_PARAM_LEVEL_STREAM)
+        {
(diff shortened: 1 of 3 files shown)


c_src/quicer_config.c
```diff
@@ -335,6 +335,13 @@ getopt3(ErlNifEnv *env, __unused_parm__ int argc,
     }
   else if (IS_SAME_TERM(eopt, ATOM_QUIC_PARAM_CONN_STATISTICS))
     {
+      if (q_ctx && Level == QUIC_PARAM_LEVEL_STREAM)
```
z8674558, line 338: what variables start with uppercase and what do not?

z8674558 (approved):

c_src/quicer_config.c
qzhuyan: There is no standard for the moment. 

I try to keep naming to follow like:

if it is borrowed from or there is a reflection in msquic, we use the same Var name.

if it is pure quicer internal var, snake name is preferred.

Nif sits in the middle of erlang code and msquic, the styling becomes a problem indeeded.

Feel free to change if something is undesired/ugly.

PR #9, opened 2021-03-16
feat(api) prepare atoms for QUIC_SETTINGS.

qzhuyan:


c_src/quicer_connection.c
```diff
@@ -399,31 +399,40 @@ sockname1(ErlNifEnv *env, __unused_parm__ int args,
...
+          enif_make_tuple8(
+            env,
+            enif_make_int(env, ntohs(ip[0])),
+            enif_make_int(env, ntohs(ip[1])),
+            enif_make_int(env, ntohs(ip[2])),
+            enif_make_int(env, ntohs(ip[3])),
+            enif_make_int(env, ntohs(ip[4])),
+            enif_make_int(env, ntohs(ip[5])),
+            enif_make_int(env, ntohs(ip[6])),
+            enif_make_int(env, ntohs(ip[7]))),
```
z8674558, line 422: why ntohs only when ipv6?

z8674558 (approved):

c_src/quicer_connection.c
qzhuyan: Ipv4 addr is 32bits long , we have 4 segs, and each one is one byte long, no reordering.
Ipv6 addr is 128 bits long, erlang ip_address() has the presentation of 8 segs tuple, each seg is two bytes long so we need reordering.

PR #17, opened 2021-03-25
feat(connect): set config thru opts

z8674558:
check if I am in a right direction. If ok, I'll add remaining client configs.

### DIFF
--- a/c_src/quicer_config.c
+++ b/c_src/quicer_config.c
@@ -154,14 +154,23 @@ ClientLoadConfiguration(ErlNifEnv *env,
   //
   // Configures the client's idle timeout.
   //
+
+  int IdleTimeoutMs = 0;
+  int PeerUnidiStreamCount = 0;
+  int PeerBidiStreamCount = 0;
+  if (!get_int_from_map(env, *option, ATOM_QUIC_SETTINGS_IdleTimeoutMs, &IdleTimeoutMs) ||
+      !get_int_from_map(env, *option, ATOM_QUIC_SETTINGS_PeerUnidiStreamCount, &PeerUnidiStreamCount) ||
+      !get_int_from_map(env, *option, ATOM_QUIC_SETTINGS_PeerBidiStreamCount, &PeerBidiStreamCount))
+    {
+      return false;
+    }
+
   Settings.IdleTimeoutMs = IdleTimeoutMs;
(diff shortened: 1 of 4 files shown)


test/quicer_SUITE.erl
```diff
@@ -537,7 +537,10 @@ simple_stream_server(Owner, Config, Port) ->
   end.
 
 default_conn_opts() ->
-  [{alpn, ["sample"]}].
+  [{alpn, ["sample"]},
+   {idle_timeout_ms, 5000},
+   {peer_unidi_stream_count, 1},
+   {peer_bidi_stream_count, 10}].
```
qzhuyan, line 543: FYI,  We need 'set socket opt' API in place to test this parm.

Test scenario:

1. start connection with peer_bidi_stream_count 1, 

2. start stream one and send data over it.
    The receiver recv the data

3. start stream two and send data over it.
    the receiver does not recv the data

4. set socket opt to peer_bidi_stream_count 10

5. receiver recv data from stream 2.

c_src/quicer_config.c
```diff
@@ -154,14 +154,23 @@ ClientLoadConfiguration(ErlNifEnv *env,
   //
   // Configures the client's idle timeout.
   //
+
+  int IdleTimeoutMs = 0;
+  int PeerUnidiStreamCount = 0;
+  int PeerBidiStreamCount = 0;
+  if (!get_int_from_map(env, *option, ATOM_QUIC_SETTINGS_IdleTimeoutMs, &IdleTimeoutMs) ||
```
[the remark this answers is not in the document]
z8674558:
```
bool get_uint64_from_map(ErlNifEnv *env, const ERL_NIF_TERM map, ERL_NIF_TERM key, uint64_t* value) {
  ERL_NIF_TERM evalue;
  if (!enif_get_map_value(env, map, key, &evalue)) {
    return false;
  }
  unsigned long value0 = 0;
  if (!enif_get_uint64(env, evalue, &value0)) {
    return false;
  }
  *value = (uint64_t)value0;
  return true;
}
```

test/quicer_SUITE.erl
z8674558: maybe my next task?

qzhuyan (approved):

PR #19, opened 2021-03-26
fix(counter): fix typing.

qzhuyan:


z8674558 (approved):

z8674558 (commented):
actually it is not typing error; I will send PR.

PR #29, opened 2021-04-01
feat(config): allow all settings as opt

z8674558:


z8674558:
I could make tests pass again by explicitly installing cmake. hehe

qzhuyan (approved):

PR #51, opened 2021-07-31
chore: change to shallow clone of msquic and all its submodules

zmstone:
My download speed is like 4MB/s:

before:
```
________________________________________________________
Executed in  188.59 secs   fish           external
   usr time  101.48 secs    1.79 millis  101.48 secs
   sys time   27.20 secs    0.28 millis   27.20 secs
```

after:
```
________________________________________________________
Executed in   53.13 secs   fish           external
   usr time   13.61 secs  713.00 micros   13.61 secs
   sys time    6.12 secs   48.00 micros    6.12 secs
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "submodules/msquic"]
-	path = submodules/msquic
-	url = https://github.com/microsoft/msquic.git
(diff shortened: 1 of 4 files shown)


qzhuyan (approved):
Good point.

I think we can try to switch to download msquic release zip file as well.

zmstone:
> Good point.
> 
> I think we can try to switch to download msquic release zip file as well.

does the zip file include openssl and googletest ?

PR #59, opened 2021-09-20
fix: handling of invalid handlers

qzhuyan:
- prevent passing invalid handlers to msquic API calls to avoid race condition
  invalid handler means the handler is either in a closing or closed state.

  The callback function of listeners/stream/connection call destroys the ctx in the background which calls `enif_release_resource` when the event of connection close/stream "shutdown  complete" is received. however the real destroy of ctx which includes closing the handler happens when the Var holding the ctx is dereferenced in erlang and gets GCed. After the handler is closed, it becomes an invalid pointer. If the nif function use the handler to call msquic API it could get a segfault.

   This change holds a reference to the ctx by calling `enif_keep_resource`  so that GC call will be delayed until `enif_release_resource` is called.

- refactor getop3
   msquic is able to bind to the correct handler in getparam call, so it is not necessary to rebind it in quicer now


include/quicer_types.hrl
```diff
@@ -80,6 +80,12 @@
 
 -type send_flags() :: non_neg_integer(). %% is sync send or not
 
+-type optlevel() :: false | %% unspecified
+                    quic_global |
+                    quic_registration |
+                    quic_registration |
```
qzhuyan, line 86: dup， need fix

test/quicer_SUITE.erl
```diff
@@ -734,26 +735,26 @@ tc_conn_controlling_process(Config) ->
...
-      {ok, 4} = quicer:send(Stm, <<"ping">>),
-      receive {quic, <<"ping">>, Stm, _, _, _} -> ok end,
-      {error, badarg} = quicer:getopt(Stm, Parm),
-      ok = quicer:close_connection(Conn),
-      SPid ! done,
-      ensure_server_exit_normal(Ref)
-  after 3000 ->
-      ct:fail("listener_timeout")
-  end.
+%% tc_getopt_raw(Config) ->
```
k32, line 738: Forgotten comment?

qzhuyan: I am unsure it is time to drop the raw (return in binary format) support so I leave it here, maybe we will add it back.

c_src/quicer_config.c
```diff
@@ -519,182 +551,62 @@ getopt3(ErlNifEnv *env,
...
-  bool isLevelOK = false;
-  bool isMalloc = false;
+  ERL_NIF_TERM elevel = argv[2];
 
   void *q_ctx;
-  void *Buffer = NULL;
-  /* QuicerListenerCTX *l_ctx = NULL; */
-  /* QuicerConnCTX *c_ctx = NULL; */
-  /* QuicerStreamCTX *s_ctx = NULL; */
+  ERL_NIF_TERM res;
```
qzhuyan, line 557: set init val

c_src/quicer_config.c
```diff
@@ -1047,3 +958,588 @@ parse_listen_on(ErlNifEnv *env, ERL_NIF_TERM elisten_on, QUIC_ADDR *Address)
...
+  else if (IS_SAME_TERM(optname, ATOM_QUIC_PARAM_CONN_IDEAL_PROCESSOR))
+    {
+      Level = QUIC_PARAM_LEVEL_CONNECTION;
+      Param = QUIC_PARAM_CONN_IDEAL_PROCESSOR;
+      // @TODO
+      res = ERROR_TUPLE_2(ATOM_STATUS(QUIC_STATUS_NOT_SUPPORTED));
+      goto Exit;
+    }
+
+  else if (IS_SAME_TERM(optname, ATOM_QUIC_PARAM_CONN_IDEAL_PROCESSOR))
```
qzhuyan, line 1112: dup, remove

c_src/quicer_config.c
```diff
@@ -1047,3 +958,577 @@ parse_listen_on(ErlNifEnv *env, ERL_NIF_TERM elisten_on, QUIC_ADDR *Address)
...
+      goto Exit;
+    }
+  else if (s_ctx->c_ctx)
+    {
+      res = get_connection_opt(env, s_ctx->c_ctx, optname, elevel);
+      goto Exit;
+    }
+  else
+    {
+      res = ERROR_TUPLE_2(ATOM_PARM_ERROR);
```
gorillainduction, line 1028: Should this be `ATOM_PARAM_ERROR`? It seems consistent throughout the code, but I wonder if it was a misspelling to begin with.

qzhuyan: It returns `{error, param_error}`

qzhuyan: It is not from msquic or OTP error code or Linux errno, must be me who intro it,  I think I get the impression from `EPERM`
Good catch I will fix it.

docs/todo.org
```diff
@@ -4,6 +4,7 @@
 
 * Features List
 Feature todo list, priority descending
+** add safty check for closed handlers
```
k32, line 7:
```suggestion
** add safety check for closed handlers
```

c_src/quicer_stream.c
```diff
@@ -316,10 +316,19 @@ async_start_stream2(ErlNifEnv *env,
...
   //
-  // note, ctx is not shared yet, thus no locking is needed.
+  // note, s_ctx is not shared yet, thus no locking is needed.
   //
   QuicerStreamCTX *s_ctx = init_s_ctx();
+
+  enif_mutex_lock(c_ctx->lock);
+  if (c_ctx->is_closed)
+    {
+      enif_mutex_unlock(c_ctx->lock);
```
k32, line 326: `destroy_s_ctx(s_ctx)` is missing?

qzhuyan: nice catch, I should move this block ahead.

c_src/quicer_config.c
qzhuyan: now I know,  it is from
```__unused_parm__``` 

but anyway `param` is better.

k32 (approved):

PR #68, opened 2021-12-27
fix: portable shebang

zmstone:


k32:
I don't think we should go out of the way to support NixOS. From my [past experience](https://github.com/NixOS/nixpkgs/issues/53834), the amount of work needed to make anything work is comparable with supporting an entire new OS.

zmstone:
it's also for consistent style.
we already have some scripts using this shebang and others are still on `/bin/bash`

thalesmg (approved):

k32 (approved):

PR #85, opened 2022-04-20
feat: make quicer:listener/1 don't crash when listener not found

zhongwencool:


qzhuyan:
LGTM.

Can I ask why for this change?

The old behavior follows the spec as in esockd to get compatibility in emqx.

https://github.com/emqx/esockd/blob/d076c917c327ad1d906a74c47c2e32af6e1e9d2d/src/esockd_listener_sup.erl#L116

zhongwencool:
> LGTM.
> 
> Can I ask why for this change?
> 
> The old behavior follows the spec as in esockd to get compatibility in emqx.
> 
> https://github.com/emqx/esockd/blob/d076c917c327ad1d906a74c47c2e32af6e1e9d2d/src/esockd_listener_sup.erl#L116

maybe we need change the esockd 's behaivor. so we don't need try catch.
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_listeners.erl#L123-L131

zhongwencool:
I want to implement this: with quicer:listener/1
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_listeners.erl#L140-L142

PS: we also need  implment: current_connection, max_connection function as esockd.
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_listeners.erl#L149

qzhuyan (approved):

Issue #91, opened 2022-05-18
unload nif module casue undefined behaviors.


qzhuyan:
When unload `quicer_nif ` erl module beam will unload the dynamic library and releases all the opened resource types. leaves undefined behavior.

We should reopen resources types in upgrade callback for reloading
```
int (*upgrade)(ErlNifEnv* caller_env, void** priv_data, void** old_priv_data, ERL_NIF_TERM load_info)
```

PR #93, opened 2022-05-31
refactor: decouple resources, no cross reference

qzhuyan:
- Handle QUIC_LISTENER_EVENT_STOP_COMPLETE for stopping the listener gracefully.
- Decouple connection and listener [major]
- Decouple connection and stream [major]
- fix crash that unload the nif module while running
- add dummy calls to init library when nif module is loaded
- make some test on macOS stable which doesn't support port/address reuse
- doc changes
- setopt/getopt needs accurate resource type as input.
- fix ret val when send is canceled. 
- drop async close stream
- drop async close connection
- conn proc could recv NST from mailbox
- use imm env for Stream
- fix: double free when stream&connection shares the same owner process

FIX: #91


src/quicer.erl
```diff
@@ -468,25 +468,30 @@ start_stream(Conn, Opts) when is_map(Opts) ->
...
 send(Stream, Data) ->
   case quicer_nif:send(Stream, Data, _IsSync = 1) of
     %% @todo make ref
     {ok, _Len} = OK ->
       receive
-        {quic, send_completed, Stream, _} ->
-          OK
+        {quic, send_completed, Stream, ?QUIC_SEND_COMPLETE_SUCCESS} ->
+          OK;
+        {quic, send_completed, Stream, ?QUIC_SEND_COMPLETE_SUCCESS} ->
```
thalesmg, line 484: Duplicated clause?

qzhuyan: Nice catch, was testing that and forget to rollback....

test/quicer_snb_SUITE.erl
```diff
@@ -846,11 +841,97 @@ tc_conn_no_gc(Config) ->
...
+                 %% We are rather testing the OTP behavior here but proves our understandings are correct.
+                 %% OTP GC callback, should not happen
+                 timeout = ?block_until(#{ ?snk_kind := debug
+                                         , context := "callback"
+                                         , function := "resource_conn_dealloc_callback"
+                                         , resource_id := CRid
+                                         , tag := "end"},
+                                        5000, 1000),
+                 timer:sleep(10000),
+                 quicer:getstat(ClientConn, [send_cnt, recv_oct, send_pend]),
```
thalesmg, line 901: Q: is this just to test it doesn't throw any errors?

qzhuyan: true, if `ClientConn` holds an invalid connection handler(already freed ), it could leads to undefined behavior, segfault...

test/quicer_snb_SUITE.erl
```diff
@@ -846,11 +841,97 @@ tc_conn_no_gc(Config) ->
...
+                 {ok, #{resource_id := CRid}}
+                   = ?block_until(#{ ?snk_kind := debug
+                                   , context := "callback"
+                                   , function := "ClientConnectionCallback"
+                                   , mark := ?QUIC_CONNECTION_EVENT_SHUTDOWN_COMPLETE
+                                   , tag := "event" },
+                                  5000, 1000),
+                 %% Give it time for gc that should not happen on var 'Conn', could be the source of flakiness.
+                 %% We are rather testing the OTP behavior here but proves our understandings are correct.
+                 %% OTP GC callback, should not happen
```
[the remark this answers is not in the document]
qzhuyan: The test is we have two references of `Conn`, one in the client process spawned in L864 called `Conn`, another in this process L873 called `ClientConn` which is sent from client process.
calling `erlang:garbage_collect/0 ` will not triggered the GC of `ClientConn` in this process since it still hold the reference at L901.

The test is to verify the terminating of client process will not trigger resource dealloc callback of `Conn` since we still have a ref in `ClientConn`

c_src/quicer_connection.c
```diff
@@ -667,32 +699,99 @@ async_connect3(ErlNifEnv *env,
...
-      destroy_c_ctx(c_ctx);
+  // Error exit, it must not be started!
+  assert(c_ctx->is_closed);
+
+  if (c_ctx->Connection)
+    { // when is opened
+
+      /*
+       We should not call *destroy_c_ctx* from here.
+       becasue it could cause race cond:
```
qzhuyan, line 758: typo `because`

zmstone (approved):

PR #96, opened 2022-06-15
feat: release binary packages

zmstone:
build and release binary packages per `OTP_VERSION-OS-ARCH`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,141 @@
+name: build
+
+on:
+  push:
+    tags:
+      - "*"
+jobs:
+  mac:
+    strategy:
+      fail-fast: false
+      matrix:
+        otp:
+          - 24.2.1-1
+          - 23.3.4.9-3
+        os:
+          - macos-11
+          - macos-10.15
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: prepare
+        run: |
+          brew update
+          brew install curl zip unzip gnu-sed kerl
+          echo "/usr/local/bin" >> $GITHUB_PATH
+          git config --global credential.helper store
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: ~/.kerl/${{ matrix.otp }}
+          key: otp-install-${{ matrix.otp }}-${{ matrix.os }}
+
+      - name: build erlang
+        if: steps.cache.outputs.cache-hit != 'true'
+        timeout-minutes: 60
+        run: |
+          kerl update releases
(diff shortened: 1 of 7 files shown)


qzhuyan (approved):

qzhuyan:
Could you do it on main branch? we will drop 0.0.9 soon.

zmstone:
> Could you do it on main branch? we will drop 0.0.9 soon.

Will merge it to main in another PR.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+          - 24.2.1-1
+        elixir:
+          - 1.13.3
+        arch:
+          - amd64
+          - arm64
+        os:
+          - ubuntu20.04
+          - ubuntu18.04
+          - ubuntu16.04
```
qzhuyan, line 78: remove ubuntu 16.04?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Create Release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: quicer ${{ github.ref }} Released
+          body: quicer ${{ github.ref }} Released
+          draft: false
+          prerelease: false
```
qzhuyan, line 136: prerelease: true

build.sh
```diff
@@ -0,0 +1,81 @@
...
+    cmake -B c_build
+    make -j "$JOBS" -C c_build
+    ## MacOS
+    if [ -f priv/libquicer_nif.dylib ]; then
+        # TODO: why cp, but not mv ?
+        cp priv/libquicer_nif.dylib "$TARGET_SO"
+    fi
+}
+
+download() {
```
qzhuyan, line 27: I prefer not to have download as default since there are lots of customization in the emqx build image that the binary built with them might not work with upstream images.

If you really want download to be default, pls `echo` loudly so developers know it is downloading from private build also have envvar for developer to have option  to compile instead.

.github/workflows/release.yaml
zmstone: we have the builder.
if there is an issue with the builder, we can have it removed.

qzhuyan (approved):

qzhuyan:
thanks for the PR. 
FYI, I also find static code check tool also need compile the code, downloading prebuild files will skip the checks.

PR #102, opened 2022-06-23
Export async connect and also retry connect

### LINKED ISSUE #101: Auto reconnects
In lossy/bad network test (20% pkts drop and 10% reorderd packets),  we found once the init handshake is lost, we are not able to reconnect.

One solution is do it in the application but maybe it is better to add support in quicer just follow what TCP stack will do with retries. 

```
# sysctl var in linux
net.ipv4.tcp_syn_retries = 6
```

qzhuyan:
fix #101

### DIFF
--- a/src/quicer.erl
+++ b/src/quicer.erl
@@ -182,9 +182,9 @@ close_listener(Listener) ->
   quicer_nif:close_listener(Listener).
 
 %% @doc
-%% Initial New Connection (Client)
+%% Initiate New Connection (Client)
 %%
-%% Initial new connection to remote endpoint with connection opts specified.
+%% Initiate new connection to remote endpoint with connection opts specified.
 %% @see async_connect/3
 %% @end
 -spec connect(inet:hostname() | inet:ip_address(),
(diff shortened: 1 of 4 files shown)


src/quicer.erl
```diff
@@ -195,26 +197,57 @@ connect(Host, Port, Opts, Timeout) when is_list(Opts) ->
...
+  {error, timeout};
+do_connect(Host, Port, Opts, Timeout, Retries) ->
+  HandshakeTOut = maps:get(handshake_idle_timeout_ms, Opts, 200),
+  RetryAfter = HandshakeTOut*Retries,
+  HandshakeTOut2 = case Timeout > RetryAfter of
+                     true ->
+                       RetryAfter;
+                     _ ->
+                       Timeout
+                   end,
```
thalesmg, line 211:
```suggestion
  HandshakeTOut2 = min(Timeout, RetryAfter),
```

thalesmg (dismissed):

thalesmg (approved):

qzhuyan:
thanks for reviews, guys!

PR #105, opened 2022-07-05
ci: add coveralls

qzhuyan:


.github/workflows/cover.yaml
```diff
@@ -0,0 +1,36 @@
...
+    strategy:
+      matrix:
+        otp:
+          - 24.3.3
+        build_type:
+          - RelWithDebInfo
+        logging:
+          - false
+        arch:
+          - x86_64
```
thalesmg, line 19: This doesn't seem to be used?

thalesmg (approved):

.github/workflows/cover.yaml
qzhuyan: Right, will update it later

PR #111, opened 2022-08-24
feat: quicer:perf_cnts()

qzhuyan:


src/quicer.erl
```diff
@@ -78,6 +78,7 @@
         , listeners/0
         , listener/1
         , controlling_process/2
+        , perf_cnts/0
```
ieQu1, line 81: cnts -> counters?

ieQu1 (approved):

PR #126, opened 2022-10-14
Added support for  custom ca certificates on both sides

jbevemyr:
Added support for cacertfile in listen and connect options, and added code to verify certificate agains the root certificates in the cacertfile.

Added test case to verify that the client only connects when the server certificate can be successfully verified agains the root ca cert in cacertfile, and the corresponding check on the server side when a client connects with a client certificate.

Added code to create cert.pem and key.pem from a CA certificate. This way they will not become outdated and it is easier to have different certificates at the client and server side (which I think is what you want in a real deployment).


c_src/quicer_config.c
```diff
@@ -333,6 +371,24 @@ load_alpn(ErlNifEnv *env,
   return true;
 }
 
+bool
+load_verify(ErlNifEnv *env,
+            const ERL_NIF_TERM *options,
+            bool default_verify)
+{
+  ERL_NIF_TERM verify_atom;
+  int verify_atom_length;
```
qzhuyan: remove, not used.

jbevemyr: It actually is used. Do you think it should be inlined? The function is used both from quicer_config.c and quicer_listen.c.

qzhuyan:
I found some solution for fixing macOS build
https://github.com/jbevemyr/quic/pull/1

c_src/quicer_connection.c
```diff
@@ -1350,7 +1400,25 @@ handle_connection_event_peer_certificate_received(
   // @TODO peer_certificate_received
   // Only with QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED set
   assert(QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED == Event->Type);
-  return QUIC_STATUS_SUCCESS;
+  // Only with QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED set
```
qzhuyan, line 1403: nit, dup with L1401

qzhuyan:
@jbevemyr we need these commits to make the CI pass.     
https://github.com/jbevemyr/quic/pull/1
Could you pick them?

jbevemyr:
Hm, not sure I understand what you want me to do...

qzhuyan:
> Hm, not sure I understand what you want me to do...

I mean cherry pick these two commits or manually add/merge them to your branch.

jbevemyr:
Sorry, I think I confused it a bit. I rebased my branch on your main and force pushed it. It should be in the merge request, right?

c_src/quicer_listener.c
```diff
@@ -47,6 +51,37 @@ ServerListenerCallback(__unused_parm__ HQUIC Listener,
...
+          X509_STORE *trusted = NULL;
+          X509_LOOKUP *lookup = NULL;
+          trusted = X509_STORE_new();
+
+          if (trusted != NULL)
+            {
+              lookup = X509_STORE_add_lookup(trusted, X509_LOOKUP_file());
+              if (lookup != NULL)
+                {
+                  if (!X509_LOOKUP_load_file(lookup,
```
qzhuyan, line 68: This will be called for every new connection,  and it is costly to call and will soon become the bottleneck

It will get improved when we get this merged to msquic.
https://github.com/microsoft/msquic/pull/3132/files.

jbevemyr: Yes, I agree. That should be solved if/when my PR to msquic is accepted.

jbevemyr: Then the cacertfile is passed on to msquic and incorporated into the SSL ctx from the start.

jbevemyr: In the meanwhile, one alternative would be to allocate the trusted structure earlier and re-use that every time instead. The only drawback of that is that it will not be updated if the contents of the cacertfile is changed.

c_src/quicer_connection.c
```diff
@@ -465,9 +476,48 @@ async_connect3(ErlNifEnv *env,
...
+                {
+                  X509_STORE_free(trusted);
+                  trusted = NULL;
+                }
+            }
+          c_ctx->trusted = trusted;
+        }
+
+      if (trusted == NULL)
+          fprintf(stderr, "failed to set cacertfile");
```
qzhuyan, line 512: Shall we prefer to return the error instead?

jbevemyr: Yes, that is probably better.

c_src/quicer_listener.c
jbevemyr: Let me know what you want me to do. Wait for the msquic change or optimize the initializtion of `trusted`.

qzhuyan: > The only drawback of that is that it will not be updated if the contents of the cacertfile is changed.
Agree, in that case We need some API to update/revoke the cacert.

jbevemyr:
Lets merge it and do fixes in follow-up.

PR #129, opened 2022-10-19
A little example program to play around with quicer API

klacke:
Hi William,
This little example can be used to provoke a number of issues, and I now think almost all of them are related to Connections that die.
Follow the run instructions and try out for for example, the sequence
```
connect.
{stream, 1}.
exit.
```

or

```
connect.
{stream, 1}.
CTL-C
```

or

```
connect.
{stream, 1}.
```
and then terminate the server

Check the process listing in each case at both the client and the server using i(). and you'll see various versions of un handled messages.

### DIFF
--- a/test/certs.erl
+++ b/test/certs.erl
@@ -1,0 +1,94 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2021 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-module(certs).
+-include_lib("kernel/include/file.hrl").
+
+
+-export([gen_ca/2,
+         gen_host_cert/3]).
+
+gen_ca(Path, Name) ->
+  %% Generate ca.pem and ca.key which will be used to generate certs
+  %% for hosts server and clients
+  ECKeyFile = filename(Path, "~s-ec.key", [Name]),
+  os:cmd("openssl ecparam -name secp256r1 > " ++ ECKeyFile),
+  Cmd = lists:flatten(
+          io_lib:format("openssl req -new -x509 -nodes "
+                        "-newkey ec:~s "
+                        "-keyout ~s -out ~s -days 3650 "
+                        "-subj \"/C=SE/O=Internet Widgits Pty Ltd CA\"",
+                        [ECKeyFile, ca_key_name(Path, Name),
+                         ca_cert_name(Path, Name)])),
+  os:cmd(Cmd).
+
+ca_cert_name(Path, Name) ->
+  filename(Path, "~s.pem", [Name]).
+ca_key_name(Path, Name) ->
+  filename(Path, "~s.key", [Name]).
+
+gen_host_cert(H, CaName, Path) ->
+  ECKeyFile = filename(Path, "~s-ec.key", [CaName]),
+  CN = str(H),
+  HKey = filename(Path, "~s.key", [H]),
+  HCSR = filename(Path, "~s.csr", [H]),
+  HPEM = filename(Path, "~s.pem", [H]),
+  HEXT = filename(Path, "~s.extfile", [H]),
+  CSR_Cmd =
+    lists:flatten(
+      io_lib:format(
+        "openssl req -new -nodes -newkey ec:~s "
+        "-keyout ~s -out ~s "
+        "-addext \"subjectAltName=DNS:~s\" "
+        "-addext keyUsage=digitalSignature,keyAgreement "
+        "-subj \"/C=SE/O=Internet Widgits Pty Ltd/CN=~s\"",
+        [ECKeyFile, HKey, HCSR, CN, CN])),
+  create_file(HEXT,
+              "keyUsage=digitalSignature,keyAgreement\n"
+              "subjectAltName=DNS:~s\n", [CN]),
+  CERT_Cmd =
+    lists:flatten(
+      io_lib:format(
+        "openssl x509 -req "
+        "-extfile ~s "
+        "-in ~s -CA ~s -CAkey ~s -CAcreateserial "
+        "-out ~s -days 500",
+        [HEXT, HCSR, ca_cert_name(Path, CaName), ca_key_name(Path, CaName),
+         HPEM])),
+  os:cmd(CSR_Cmd),
+  os:cmd(CERT_Cmd),
+  file:delete(HEXT).
+
+filename(Path, F, A) ->
+  filename:join(Path, str(io_lib:format(F, A))).
+
+str(Arg) ->
+  binary_to_list(iolist_to_binary(Arg)).
+
+create_file(Filename, Fmt, Args) ->
+  {ok, F} = file:open(Filename, [write]),
+  try
+    io:format(F, Fmt, Args)
+  after
+    file:close(F)
+  end,
+  ok.
+
+
+%%%_* Emacs ====================================================================
+%%% Local Variables:
+%%% erlang-indent-level: 2
+%%% End:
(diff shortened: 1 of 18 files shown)


klacke:
Ahh, damn, got the wrong version for you. Wait a wihile ...

klacke:
Now we're good

klacke:
Sorry for wasting your time. Just checked some small stuff that actually makes the example work.

qzhuyan:
> Sorry for wasting your time. Just checked some small stuff that actually makes the example work.

No worry, they are easy fixed.

about 
> Check the process listing in each case at both the client and the server using i(). and you'll see various versions of un handled messages.

I think for unhanded messages some of them are expected, if these messages are not interested, we could mask (PR later) them then callback will not send to the owner. 

I have fixed a issue that stream acceptor should recv a msg while the connecion is down so it will not keep waiting there. in #131. 
I will make further changes for the connection acceptors as well.

test/example/qt.erl
```diff
@@ -0,0 +1,402 @@
...
+
+
+acceptor(Top, L) ->
+    io:format("Call accept ~p\n", [self()]),
+    {ok, Conn} = quicer:accept(L, [], infinity),
+
+    Top ! {self(), more},
+
+    io:format("accept -> ~p\n",[Conn]),
+
```
qzhuyan, line 77: FYI, maybe call `quicer_nif:async_accept_stream` here, force we have stream acceptor in the wait list before handshake is done.

Because there could be a race here that when new stream is created by remote (client) (after handshake) we dont have acceptor yet as in L89 hasn't been called. The quicer will select connection owner as the stream owner as fallback.

test/example/qt.erl
```diff
@@ -0,0 +1,402 @@
...
+            ok = quicer:setopt(S, active, 20),
+            recv_ping(S, N, CNo, SNo);
+        {quic, <<"ping">>, S,_} ->
+            io:format("Got ping ~p from ~p:~p\n",[N, CNo, SNo]),
+            {ok, 4} = quicer:send(S, <<"pong">>),
+            recv_ping(S, N+1, CNo, SNo);
+        {quic, <<"die">>, S,_} ->
+            io:format("Got die from ~p:~p\n",[CNo, SNo]),
+            die;
+        {quic, peer_send_shutdown, S} ->
```
qzhuyan, line 137: In latest main branch, we use 4 elements tuple.  so should be 
```
{quic, peer_send_shutdown, S, _} ->
```

klacke:
The sequence connect. {stream, 1}. exit. leaves two processes with unhandled messages. 
1. quicer:accept_stream() hangs, when it should return an error with 2 mesages in the queue
```
{messages,[{quic,shutdown,#Ref<0.798538594.392822784.31685>,
                  success},
            {quic,closed,#Ref<0.798538594.392822784.31685>,
                  #{is_app_closing => false,is_handshake_completed => true,
                    is_peer_acked => false}}]},
```

2. A process hangs indefintely in quicer:shudown_stream with a message in the queue


```

File: test/example/qt.erl (hunk 2 above)
Comment (klacke) [reply test/example/qt.erl]: Yes, indeed, fixed, pushed

**Comment by klacke** [comment]
Have a look at https://github.com/hawk/lux it's a test tool that we're using extensively. If you want I can add lux tests to this example, testing for example things like the client/server behaving odd.  3 different tests make sense for both the client and the server. Beam suspends, Beam exits, the erlang process holding the stream exits.

**Comment by klacke** [comment]
Just say no if you don't want it

**Comment by qzhuyan** [comment]
> Have a look at https://github.com/hawk/lux it's a test tool that we're using extensively. If you want I can add lux tests to this example, testing for example things like the client/server behaving odd. 3 different tests make sense for both the client and the server. Beam suspends, Beam exits, the erlang process holding the stream exits.

Yes pls! that will be helpful!

**Comment by qzhuyan** [comment]
The sequence connect. {stream, 1}. exit. leaves two processes with unhandled messages.

    quicer:accept_stream() hangs, when it should return an error with 2 mesages in the queue

{messages,[{quic,shutdown,#Ref<0.798538594.392822784.31685>,
                  success},
            {quic,closed,#Ref<0.798538594.392822784.31685>,
                  #{is_app_closing => false,is_handshake_completed => true,
                    is_peer_acked => false}}]},


These two messages  are for connection owner, not for stream owner. We cannot handle these two in `quicer:accept_stream` because stream owner and connection could be different processes.  ` quicer:accept_stream()` now expecting `{error, closed}` in PR #131 Maybe we could merge that PR first.

**Comment by qzhuyan** [comment]
about 
>    A process hangs indefintely in quicer:shudown_stream with a message in the queue
what is the message in the process mailbox? If it is for 'connection closed' `{quicer, closed, Ref, #{} `}, I really want script with lux to reproduce.  because we *MUST* get a `{quic, stream_down, StmRef, _}`  when connection is closed.

**Comment by klacke** [comment]
```
 {messages,[{quic,send_shutdown_complete,
                  #Ref<0.798538594.392822784.31687>,false}]},
 {current_stacktrace,[{quicer,shutdown_stream,4,
                              [{file,"/home/klacke/git/quic/src/quicer.erl"},
                               {line,636}]},
                      {qt,recv_ping,4,[{file,"qt.erl"},{line,133}]},
                      {qt,stream_owner,2,[{file,"qt.erl"},{line,107}]},
                      {proc_lib,init_p,3,
                                [{file,"proc_lib.erl"},{line,225}]}]}]
```

**Comment by klacke** [comment]
I'll get started on the lux tests ASAP

**Comment by klacke** [comment]
OTOH, if the connection owner actually got the closed message as opposed to the stream owner getting it, it would be the job of the connection owner to shutdown all spawned stream owners.

**Comment by qzhuyan** [comment]
> OTOH, if the connection owner actually got the closed message as opposed to the stream owner getting it, it would be the job of the connection owner to shutdown all spawned stream owners.

From my testing, stream owners will get stream-down messages  (due to connection close). We don't really need to link connections and streams in the erlang world, they could run independently.

**Comment by qzhuyan** [comment]
> {messages,[{quic,send_shutdown_complete,
>                   #Ref<0.798538594.392822784.31687>,false}]},
>  {current_stacktrace,[{quicer,shutdown_stream,4,
>                               [{file,"/home/klacke/git/quic/src/quicer.erl"},
>                                {line,636}]},
>                       {qt,recv_ping,4,[{file,"qt.erl"},{line,133}]},
>                       {qt,stream_owner,2,[{file,"qt.erl"},{line,107}]},
>                       {proc_lib,init_p,3,
>                                 [{file,"proc_lib.erl"},{line,225}]}]}]

Just clarify, looks like it is blocking for peer node from closing the stream at its end *OR* connection idle timeout (in your case 9s timeout).
Either case should give us a stream close as following and unblock the shutdown_stream
```{quic, stream_closed, stream_handler(), _  }```

**Comment by klacke** [comment]
Hi William, I'm not still sure you want this PR, nevertheless, it was necessary for me to ensure that quic could be used in a manner that worked for us. I've pushed to my fork, various fixes and as far as I can see, everything works perfectly now. I added a bunch of lux tests that check various odd scenarios including CTL Z suspend of both client and server ensuring all streams and connections are cleared.
So if you want this, I'm ready for a merge now.

**Comment by klacke** [comment]
Thus, if you want to - you can close all the issues I created, they're all moot now as far as I can see.

**Comment by qzhuyan** [comment]
I am just waiting for a "ping" from you. I will play with it. thanks for the PR!

File: test/certs.erl [hunk 3]
```diff
@@ -0,0 +1,94 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2021 EMQ Technologies Co., Ltd. All Rights Reserved.
```
qzhuyan, line 2: 2020-2022

test/certs.erl
```diff
@@ -0,0 +1,94 @@
...
+%%
```
qzhuyan, line 3: nit, could we rename it to `quicer_test_lib_certs`?

qzhuyan: maybe just `quicer_test_lib`

klacke:
Absolutely, will fix. Found some other smallish issues too.

klacke:
Hi William, just pushed according to your req, and also and also added a
few more tests, ping1000.lux fails and there seems to be a new problem with
quicer:setopt(S, active, N).
This used to work before, I rebased from emxq main before testing.

Cheers  /klacke

qzhuyan:
@klacke thanks for the update, I am going to merge it.

for the issue of set active, is that a Server accepted stream?

klacke:
It is server accepted stream, qt.erl hangs at
```
{messages,[]},
 {current_stacktrace,[{qt,recv_ping,4,
                          [{file,"qt.erl"},{line,124}]},
                      {qt,stream_owner,2,[{file,"qt.erl"},{line,116}]},
                      {proc_lib,init_p,3,
                                [{file,"proc_lib.erl"},{line,225}]}]}]
```

and last printout is `Setting active 20`

qzhuyan:
@klacke should be fixed in #138

PR #135, opened 2022-10-27
feat: mark orphan stream

qzhuyan:
orphan stream is marked in new_stream props

Old:
```
{quic, new_stream, stream_handler(), stream_open_flags()}
```

New:
```
{quic, new_stream, stream_handler(), #{ flags := stream_open_flags()
                                      , is_orphan := boolean()
                                      }}
```


c_src/quicer_connection.c
```diff
@@ -1219,11 +1222,17 @@ handle_connection_event_peer_stream_started(QuicerConnCTX *c_ctx,
...
   s_ctx->is_closed = FALSE;
 
-  ERL_NIF_TERM report
-      = make_event(env,
-                   ATOM_NEW_STREAM,
-                   enif_make_resource(env, s_ctx),
-                   enif_make_uint(env, Event->PEER_STREAM_STARTED.Flags));
+  ERL_NIF_TERM props_name[] = { ATOM_FLAGS, ATOM_IS_ORPHAN };
+  ERL_NIF_TERM props_value[]
+      = { enif_make_int(env, Event->PEER_STREAM_STARTED.Flags),
```
thalesmg, line 1227: Just confirming: this was a `make_uint` before.  Is this change to `make_int` intentional?

ieQu1 (approved):

c_src/quicer_connection.c
qzhuyan: good catch, I don't recall there was a reason to change it.  we only use 4bits should be fine.
But it is just nice to use uint. will fix that!

PR #141, opened 2022-11-08
Some sort of timing bug found, the sleep statement at line 222 in qt.…

klacke:
…erl seems to alleviate it.

If an active, false stream is created directly after connect, sometimes the server side still thinks its in active true mode. Timing related, works approx 50 % of the time.

### DIFF
--- a/test/example/qt.erl
+++ b/test/example/qt.erl
@@ -90,28 +90,24 @@ accept_stream_loop(Conn) ->
   case quicer:accept_stream(Conn, [{active, false}], infinity) of
     {ok, Stm} ->
       io:format("accept stream -> ~p\n",[Stm]),
-
       Self = self(),
       H = proc_lib:spawn_link(
             fun() -> stream_owner(Self, Stm) end),
       ok = quicer:controlling_process(Stm, H),
-
-      receive
-        {H, continue} ->
-          accept_stream_loop(Conn)
-      end;
-
+      H ! {self(), continue},
+      accept_stream_loop(Conn);
     Err ->
       io:format("Failed to accept_stream ~p leaving accept stream\n",[Err]),
       ok
   end.
 
 stream_owner(Top, Stream) ->
-  Top ! {self(), continue},
+  receive {Top, continue} ->
+      ok
+  end,
   case server_negotiate(Stream) of
     {ok, CNo, SNo} ->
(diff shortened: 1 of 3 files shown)


test/example/qt.erl
```diff
@@ -90,28 +90,24 @@ accept_stream_loop(Conn) ->
   case quicer:accept_stream(Conn, [{active, false}], infinity) of
```
qzhuyan, line 90: The race is before this line get called,  you get incoming stream,  this is what we called 'orphan stream'. orphan stream inherits active mode of the connection owner.


try change L73 to 
```
{ok, Conn} = quicer:accept(L, [{active,false}], infinity),
```

also see doc:
https://github.com/emqx/quic/blob/main/docs/messages_to_owner.md?plain=1#L279

another way to avoid race is to call async_accpet_stream before handshake L79 and then wait for the `{quic, new_stream, ... }` message.

BTW, I will fix the example in README.

klacke: Hmm, tried the 
```
-  {ok, Conn} = quicer:accept(L, [], infinity),
+  {ok, Conn} = quicer:accept(L, [{active, false}], infinity),
```

and still get `{error,einval}` when quicer:recv() is called on an active stream.
In this type of code, you really want active false, ensuring that no application data is ever sent to the parent. Alternatively, if `{active, false}` isn't possible in combination with spawn + controlling_process + sync and continue, the first data sent must come from the server side.

I understand the orphan thing, I guess it's similar to the `LINGER` option to UNIX `listen()`

As for calling `async_accept_stream` I'm not sure I understand what you mean, what you suggest is exactly what  `accept_stream()` does, i.e call `async_accept_stream()` and then wait for the message. 

As for the commenting and and timeout, @overudp , even `{connect_stream, 1}` fails sometimes.

qzhuyan: @klacke I see . thanks for explanation. 

I already had a fix for this issue, the solution is to set orphan stream to passive by default. It is on my multi stream support branch, not on main branch yet.

You may try to cherry pick the commit from my fork: 
https://github.com/qzhuyan/quic/commit/fea570c447fdf510a3f03aa12e0f0b9e5a4ea14c

klacke:
Excellent, I'll wait until merged.

Cheers

PR #143, opened 2022-11-10
Klacke/example nat site

klacke:
Hi William, I've stumbled upon more errors. Take a look at rev.erl  and nat.lux in this branch.

    The server does the initial accept() and following that
    The server calls start_stream(), and the original client calls
    accept_stream()
    
    Two possible bugs found with this example.
    
    1. The call to close on line 61 doesn't seem to go through
       to the other end.
    
    2. Look at the lux file, if the test  is run twice it doesn't work.

### DIFF
--- a/test/example/rev.erl
+++ b/test/example/rev.erl
@@ -1,0 +1,137 @@
+-module(rev).
+-compile(export_all).
+-compile(nowarn_export_all).
+
+-include("../../include/quicer.hrl").
+
+%% This is an axample of how we can revert connections. This is especially
+%% interesting when we want to connect to a device behind a NAT device.
+%% Code behind the NAT device can do connect(), and following that the
+%% code at the "top site" can do start_stream() to the device behind the NAT
+
+%% To run:
+
+%% > Top = rev:start().
+%% > Top ! connect.
+%% > Top ! connect.
+
+
+-define(INTERVAL, 3000).
+
+
+
+start() ->
+    application:ensure_all_started(quicer),
+    _C = proc_lib:spawn_link(fun() ->
+                                     nat_site()
+                             end),
+
+    Top = proc_lib:spawn_link(fun() ->
+                                      listener()
+                              end),
+    Top.
+
+listener() ->
+    Port = 4567,
+    LOptions = [ {cert, "./server.pem"}
+               , {key,  "./server.key"}
+               , {verify, none}
+               , {active, false}
+               , {handshake_idle_timeout_ms, 3 * ?INTERVAL}
+               , {keep_alive_interval_ms, ?INTERVAL}
+               , {alpn, ["sample"]}
+               , {idle_timeout_ms, 3 *?INTERVAL}
+               , {peer_bidi_stream_count, 64000}
+               ],
+
+    {ok, L} = quicer:listen(Port, LOptions),
+    {ok, Conn} = quicer:accept(L, [{active, false}], infinity),
+    {ok, Conn} = quicer:handshake(Conn),
+    r_loop(Conn).
+
+r_loop(Conn) ->
+    receive
+        connect ->
+            io:format("Starting stream downwards \n",[]),
+            {ok, S} = quicer:start_stream(Conn, [{active, false}]),
+            ok = quicer:setopt(S, active, 20),
+            ok = send_ping(5, S),
+            io:format("Closing stream \n",[]),
+            %ok = quicer:close_stream(S),
+            ok = quicer:async_close_stream(S),
+            io:format("Stream closed \n",[])
+    end,
+    r_loop(Conn).
+
+
+nat_site() ->
+    Port = 4567,
+    case quicer:connect("localhost", Port,
+                        [{alpn, ["sample"]},
+                         {verify, none},
+                         {keep_alive_interval_ms, ?INTERVAL},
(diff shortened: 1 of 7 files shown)


qzhuyan:
@klacke thanks for the PR, do you see any chance to make these example test run in GitHub actions? 
Would be great if we could have them in the CI pipeline.

klacke:
I'll look into that if you want, we're certainly running lux scripts in our
CI/CD pipeline on gitlab

klacke:
Uhh, github workflows this is not really my strong side, I bet you are much
faster than me with this, after all you wrote main.yml
What I'll do instead is polish the examples until they shine, I have one
more slightly more complex example in
the pipeline. It doesn't work yet - and the last one with nat.lux is a
simplification of that example.
I'm trying to get something that is as close to our use case as possible.
The use case is a cluster we call top which
has a LB in front of it, typically in AWS and then a very large number of
edge sites that "call home". They may or may not
be behind a NAT. Code running at the top site needs to initiate connections
to the edge sites, thus those connections must be initiated from the edge
and then turned around.

test/example/rev.erl
```diff
@@ -0,0 +1,137 @@
...
+    r_loop(Conn).
+
+
+nat_site() ->
+    Port = 4567,
+    case quicer:connect("localhost", Port,
+                        [{alpn, ["sample"]},
+                         {verify, none},
+                         {keep_alive_interval_ms, ?INTERVAL},
+                         {handshake_idle_timeout_ms, 3 * ?INTERVAL},
```
qzhuyan, line 73: add 
```
 {peer_bidi_stream_count, 5},
```
default is one.

test/example/rev.erl
```diff
@@ -0,0 +1,137 @@
...
+    {ok, L} = quicer:listen(Port, LOptions),
+    {ok, Conn} = quicer:accept(L, [{active, false}], infinity),
+    {ok, Conn} = quicer:handshake(Conn),
+    r_loop(Conn).
+
+r_loop(Conn) ->
+    receive
+        connect ->
+            io:format("Starting stream downwards \n",[]),
+            {ok, S} = quicer:start_stream(Conn, [{active, false}]),
```
qzhuyan, line 56: This call hides the error when peer is flow controlled on the number of streams. 
you can use  open flag `QUIC_STREAM_START_FLAG_FAIL_BLOCKED` to fail the start instead.
https://github.com/emqx/quic/blob/main/include/quicer_types.hrl#L139

The default behavior is waiting for peer to remove the limit , thus long blocking here.

qzhuyan:
The test works if you rebase on to my branch and also apply the remarks above(bump stream limit): https://github.com/emqx/quic/pull/144 

check it here:

https://github.com/qzhuyan/quic/tree/klacke/example_nat_site

klacke:
Fantastic, and the close goes through too. All good :-)

klacke:
Hmm, I was wrong, the close from the top doesn't go through. The stream_handler/2 is waiting for `stream_closed` message but all it gets is a `peer_send_shutdown` message.
```
 {messages,[{quic,peer_send_shutdown,
                  #Ref<0.2767721431.4258922496.17127>,undefined}]},
 {current_stacktrace,[{rev,stream_handler,2,
                           [{file,"rev.erl"},{line,104}]},
                      {rev,stream_handler0,1,[{file,"rev.erl"},{line,99}]},
                      {proc_lib,init_p,3,
                                [{file,"proc_lib.erl"},{line,225}]}]}]

```

klacke:
Thus

```
diff --git a/test/example/rev.erl b/test/example/rev.erl
index 4568897..62e6f64 100644
--- a/test/example/rev.erl
+++ b/test/example/rev.erl
@@ -57,8 +57,8 @@ r_loop(Conn) ->
             ok = quicer:setopt(S, active, 20),
             ok = send_ping(5, S),
             io:format("Closing stream \n",[]),
-            %ok = quicer:close_stream(S),
-            ok = quicer:async_close_stream(S),
+            ok = quicer:close_stream(S),
+            %ok = quicer:async_close_stream(S),
             io:format("Stream closed \n",[])
     end,
     r_loop(Conn).
```

Makes the test hang

qzhuyan:
>Hmm, I was wrong, the close from the top doesn't go through in your branch. The stream_handler/2 is waiting for > stream_closed message but all it gets is a peer_send_shutdown message.


the shutdown_stream API use *graceful* shutdown by default, thus it waits for shutdown ack from the remote otherwise it will wait for idle timeout on the connection level.

I guess adding the handling at the `NAT` side should solve the issue.  something like 
```
%% in the NAT recv loop
receive {quic, peer_send_shutdown, S, _X} -> quicer:shutdown_stream(S) end.
```

It could happen that the remote  (NAT) do an abortive shutdown, so I think it is good idea to use async_shutdown stream to handle all the situations (graceful, abortive)

qzhuyan:
@klacke are you going to update this branch or we could merge?

klacke:
Hi William, I'm down with covid, but almost out of the fog today, I'll take
a look later today and get back to you.

Cheers

klacke:
Yes, ready to merge. I cleaned up the example and `lux .` works ok now.

qzhuyan (approved):

PR #149, opened 2022-12-07
Enhanced the rev.erl code, sharing multiple streams on the same conne…

klacke:
…ction

I enhanced the rev.erl code making it more interesting. I also added a test and a README.md file


klacke:
Question:  I should be using the handoff code as well, shouldn't I

klacke:
Yes, I see now, quicer:handoff_stream/3 should be used, 
I think that maybe quicer:controlling_process should do what handoff_stream does, instead of a new function. It's also confusing to still export controlling_process if it shouldn't be used. OTOH, I guess it stlll should be used for Connections, yes ?

klacke:
I'll change the code to use handoff_stream instead

qzhuyan:
thanks for the PR! 
>I think that maybe quicer:controlling_process should do what handoff_stream does, instead of a new function. It's also confusing to still export controlling_process if it shouldn't be used. OTOH, I guess it stlll should be used for Connections, yes ?

thanks for the suggestion, yes `controlling_process` could be used for both streams and connections.
We need handoff_connecion for connection, too.

qzhuyan (approved):

PR #154, opened 2022-12-15
chore: enable OTP-25 builds

keynslug:


.github/workflows/cover.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         otp:
-          - 24.3.3
```
qzhuyan, line 13: we want to keep OTP24.

.github/workflows/main.yml
```diff
@@ -9,15 +9,15 @@ jobs:
         os:
           - macos-11
         otp:
-          - 24
```
qzhuyan, line 12: we want to keep 24, is it still possible?

qzhuyan (approved):

.github/workflows/main.yml
keynslug: Sure. Brought it back.

qzhuyan (approved):

PR #156, opened 2022-12-15
fix(conn): flush closed event when sync conn fail

### LINKED ISSUE #150: quic:connect()
If quic:connect() fails nicely with for example 
```
{error,transport_down,
                            #{error => 1,status => unreachable}}
```

an unnecessary  message

```
{quic,closed,#Ref<0.1073868949.2575171596.191148>,
            #{is_app_closing => false,is_handshake_completed => false,
              is_peer_acked => true}}
```

is subsequently sent to the caller. 

I find this weird, the call has already failed. Also the flag `is_peer_acked` is confusing.

qzhuyan:
fix #150

When sync conn API call is failed after conn start, the conn closed event is useless to the caller because the handle
will not be presented to the caller.  the closed event is just non interesting junk in the proc mailbox.

This PR flush it before returns to the caller.

No changes in NIF because for client, no conn closed event will be sent if handshake is failed.

### DIFF
--- a/src/quicer.erl
+++ b/src/quicer.erl
@@ -216,8 +216,10 @@ connect(Host, Port, Opts, Timeout) when is_map(Opts) ->
           {ok, H};
         {quic, transport_shutdown, H, Reason} when Reason == connection_timeout
                                                    orelse Reason == connection_idle ->
+          flush(closed, H),
           {error, timeout};
-        {quic, transport_shutdown, _, Reason} ->
+        {quic, transport_shutdown, H, Reason} ->
+          flush(closed, H),
           {error, transport_down, Reason}
       end;
     {error, _} = Err ->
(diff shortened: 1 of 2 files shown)


zmstone (approved):

src/quicer.erl
```diff
@@ -951,6 +953,14 @@ do_forward_stream_msgs(Stream, Owner, MRef) ->
             ok
     end.
 
+%% @doc garbage collect some quic event that is useless to the caller.
+-spec flush(atom(), handle()) -> ok.
+flush(QuicEventName, Handle) when is_atom(QuicEventName) ->
+  receive
+    {quic, QuicEventName, Handle, _} -> ok
```
zmstone, line 960: if it should have been delivered already,
maybe raise an error exception instead of blocking forever.

qzhuyan: The event is sent from NIF code, not from another process.
It is like an assertion too, the event must be in the mailbox or comes late from NIF.

yes maybe have some reasonable timeout and raising exceptions is much nicer than blocking forever.

PR #164, opened 2023-01-11
Dialyzer work on test/example

klacke:
Found a few type errors in the quicer code as well. Especially, the cacert stuff must be part of listen opts for the server

### DIFF
--- a/test/example/qt.erl
+++ b/test/example/qt.erl
@@ -44,15 +44,15 @@
 s() ->
   application:ensure_all_started(quicer),
   Port = 4567,
-  LOptions = [ {cert, "./server.pem"}
-             , {key,  "./server.key"}
-             , {verify, none}
-             , {handshake_idle_timeout_ms, 3 * ?INTERVAL}
-             , {keep_alive_interval_ms, ?INTERVAL}
-             , {alpn, ["sample"]}
-             , {idle_timeout_ms, 3 *?INTERVAL}
-             , {peer_bidi_stream_count, 64000}
-             ],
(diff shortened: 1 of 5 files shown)


klacke:
I'm not quite sure what the rebar3 dialyzer call at the top Makefile actually does, but it's clearly not running dialyzer on all the code, especially not the test code.

klacke:
To run dialyzer on the examples test code, I did, `make all` in `test/example` and then
`dialyzer -r .` at the top

test/example/qt.erl
```diff
@@ -70,14 +70,14 @@ listener(L) ->
...
-  {ok, Conn} = quicer:accept(L, [], infinity),
+  {ok, Conn} = quicer:accept(L, #{}, infinity),
 
   Top ! {self(), more},
 
   io:format("accept -> ~p\n",[Conn]),
 
   case quicer:handshake(Conn) of
-    {ok, Conn} ->
+    {ok, _Conn} ->
```
qzhuyan, line 80: why this change?

klacke: Sigh, stupid, just  a misdirected attempt at getting through dialyzer, skip that

qzhuyan:
I am not quiet sure if we need to run dialzyer over test code in the CI (at least not that urgent) but feel free to make a PR if you think it is necessary.

klacke:
Thx.
As for dialyzer on the test code, since the test code is one of the primary sources for learning your lib, I feel it's important that the test code use the library the way it's intended to be used. I guess it's mostly an issue of the various options being a map as opposed to key,val list.

PR #168, opened 2023-01-19
fix resource leaks in test

qzhuyan:
mem check tool reports some indirect mem leak due to

- some owner process is not stopped in some testcase
- some connection isn't aborted after test
- some uninterested signals from quic are not consumed from mailbox of CT process.

### DIFF
--- a/test/quicer_test_lib.erl
+++ b/test/quicer_test_lib.erl
@@ -19,7 +19,9 @@
 
 
 -export([gen_ca/2,
-         gen_host_cert/3]).
+         gen_host_cert/3,
+         receive_all/0
+        ]).
 
 gen_ca(Path, Name) ->
   %% Generate ca.pem and ca.key which will be used to generate certs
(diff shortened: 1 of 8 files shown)


test/quicer_test_lib.erl
```diff
@@ -87,6 +89,16 @@ create_file(Filename, Fmt, Args) ->
...
 
+receive_all() ->
+  receive_all([]).
+
+receive_all(Res)->
+  receive
+    X ->
+      receive_all([X|Res])
+  after 0 ->
+      Res
```
thalesmg, line 100: Tiny nit: maybe reverse the list to keep it in chronological order?

thalesmg (approved):

thalesmg (approved):

test/quicer_test_lib.erl
qzhuyan: that will be helpful indeed.

PR #172, opened 2023-01-20
fix(resource_down):  double check ownership

qzhuyan:
After controlling_process/2 call, the old owner is demonitored *asynchronously*. That is we may get a stream/conn resource_down callback with the pid of old owner while the new ower is already set in the resource ctx.

This change double check if the resource owner is the dead process pid and shutdown the resource only when they match.

### DIFF
--- a/test/quicer_SUITE.erl
+++ b/test/quicer_SUITE.erl
@@ -1199,6 +1199,7 @@ tc_stream_controlling_process_demon(Config) ->
         {'DOWN', MonRef, process, NewOwner, {Res, Stm}} ->
           ct:pal("Set controlling_process res: ~p", [Res])
       end,
+      ?assertEqual({error, owner_dead}, quicer:controlling_process(Stm, NewOwner)),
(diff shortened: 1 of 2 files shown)


test/quicer_SUITE.erl
```diff
@@ -1177,6 +1180,48 @@ tc_stream_controlling_process(Config) ->
...
+      ok = quicer:setopt(Stm, active, true),
+      {ok, _Len} = quicer:send(Stm, <<"owner_changed">>),
+      receive
+        {quic, <<"owner_changed">>, Stm, _} ->
+          ok
+      end,
+      %% Set controlling_process again
+      {NewOwner2, MonRef2} = spawn_monitor(fun() -> receive stop -> ok end
+                                           end),
+      %?assertEqual({error, owner_dead}, quicer:controlling_process(Stm, NewOwner)),
```
qzhuyan, line 1211: This cause issue: set owner to dead pid will return error but makes the stream has no owner.

zmstone (approved):

PR #173, opened 2023-01-23
feat: future packet buffer

qzhuyan:
Sometimes we find msquic deliver data to app out of order.
This happens when system run out of memory for a short while and
then recover OR the data is lost and never detect the loss or recover.

We need this extra guard to ensure the order by using the absolute
offset from the quic event. We prefer stuck the stream/connection rather
than providing wrong data to the APP layer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,18 +27,18 @@ xref:
 	$(REBAR) xref
 
 .PHONY: eunit
-eunit: compile
-	$(REBAR) eunit verbose=truen
+eunit:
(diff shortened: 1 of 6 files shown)


zmstone (approved):

nibanks:
Just curious @qzhuyan, can you explain this statement?
> Sometimes we find msquic deliver data to app out of order.
For a single stream, MsQuic should never deliver data out of order. Across streams, all bets are off, because they are independent units.

PR #179, opened 2023-02-22
feat: release binaries for rocky linux 9 (el9)

id:
also deprecate macos 10.15 and upgrade some actions


.github/workflows/release.yaml
```diff
@@ -86,6 +85,7 @@ jobs:
           - debian9
           - el7
           - el8
+          - el9
```
thalesmg, line 88: Q: should add `amzn2` as well?

thalesmg (approved):

qzhuyan (approved):

PR #181, opened 2023-03-02
Added a variant of qt to compare quic to SSL

klacke:


test/example/README.md
```diff
@@ -45,6 +45,10 @@ ok
 Also, it's interesting is to experiment with various variants of CTL-Z and
 CTL-C on both the server and the client.
 
+Furthermore, there is an almost identical variant called qt_ssl.erl that can
+be used to perform comaparisions between quic and SSL. Both speed and memory
+consumption tests can be performed. Spoiler, quic is both smaller and faster
+than OTP SSL.
```
qzhuyan, line 51: good to know!
We also did some benchmarks in early stage last year
https://www.emqx.com/en/blog/mqtt-over-quic

Some info is outdated, and we will update them after the major EMQX 5.0 major release.

test/example/qt.erl
```diff
@@ -457,6 +484,28 @@ check_processes() ->
...
+perf2(S, N, Sz, Packet) ->
+  {ok, Sz} = quicer:send(S, Packet),
+  {ok, Bin} = quicer:recv(S, Sz),
+  Sz = size(Bin),
+  perf2(S, N-1, Sz, Packet).
+
+
+
+pack(N) ->
+  list_to_binary(
```
qzhuyan, line 504: Wonder if you map all values to `18`? do you meet issue with incomplete/corrupted data?
I am asking to ensure you don't get the above issues.


nit, if you want an increasing integers binary stream,  here is a simplified version.
```
<< << A >> || A <- lists:seq(1, N) >>.
```

qzhuyan (approved):

PR #183, opened 2023-03-03
Added a longrunning test that never ends, can be used to inspect memo…

klacke:
…ry characteristics.

### DIFF
--- a/test/example/rev.erl
+++ b/test/example/rev.erl
@@ -31,6 +31,12 @@
 
 -define(INTERVAL, 3000).
 
+%% Uncomment iof when running longrun() in order to be able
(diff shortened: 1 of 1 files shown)


klacke:
very good news here, running for hours, indicates zero memory loss.

qzhuyan:
> very good news here, running for hours, indicates zero memory loss.

thanks!
FYI, we did 3 weeks long duration tests, 20k sends/s from 2000 clients to three servers,  no mem leak. 
But the test scenario is mostly happy path, we are testing more on negative test scenarios.

qzhuyan (approved):

PR #198, opened 2023-07-03
feat(listener): close with timeout

qzhuyan:


c_src/quicer_listener.c
```diff
@@ -473,13 +474,19 @@ close_listener1(ErlNifEnv *env,
...
+  if (l_ctx->is_closed)
+  {
+    ret = ERROR_TUPLE_2(ATOM_CLOSED);
+  }
   l_ctx->is_closed = TRUE;
   enif_mutex_unlock(l_ctx->lock);
 
   // It is safe to close it without holding the lock
   // This also ensures no ongoing listener callbacks
-  // This is a blocking call. @TODO have async version or use dirty scheduler
```
paulozulato, line 481: was the *TODO* removed on purpose or by accident? (as it's not related to this change)

c_src/quicer_listener.c
```diff
@@ -473,13 +474,19 @@ close_listener1(ErlNifEnv *env,
...
+  // This is a blocking call.
+  //
   MsQuic->ListenerClose(l);
```
paulozulato, line 489: is it OK to close the listener if it was already closed? Or should it be better to return just after the critical section if ret != ATOM_OK?

qzhuyan: It is ok. 
L476 set it to NULL and ListenerClose has `_Pre_defensive_` annotation. So it will never trigger use-after-free.

c_src/quicer_listener.c
qzhuyan: We could use ListenerStop for async calls instead and keep ListenerCLose sync. 
maybe it is a good time to add ListenerStop support. 
I will update the PR.

src/quicer.erl
```diff
@@ -200,9 +200,28 @@ listen(ListenOn, Opts) when is_map(Opts) ->
...
-  quicer_nif:close_listener(Listener).
+  close_listener(Listener, 5000).
+
+-spec close_listener(listener_handle(), timer:time()) ->
+        ok | {error, badarg | closed | timeout}.
+close_listener(Listener, Timeout) ->
+  case quicer_nif:close_listener(Listener) of
+    ok ->
+      receive
+        {quic, listener_stopped, Listener} ->
```
qzhuyan, line 213: The event is only shot to the Listener owner who opens the listener. 
The timeout prevents long blocking on the caller who is not the owner.
maybe we should check the owner ship in NIF and return `{error, not_owner}`

qzhuyan:
replaced with #205

PR #205, opened 2023-08-22
feat: stop then start listener

qzhuyan:
Improve listener API,
- add API to stop and restart the listener with Listener handle
- fix config resource leaks in listener
- rename start_listener to spawn listener for supervised listener
- rename stop_listener to terminate listener for supervised listener

### DIFF
--- a/c_src/quicer_listener.c
+++ b/c_src/quicer_listener.c
@@ -216,15 +216,6 @@ ServerListenerCallback(__unused_parm__ HQUIC Listener,
 
     case QUIC_LISTENER_EVENT_STOP_COMPLETE:
       env = l_ctx->env;
-
-      // Close listener in NIF CTX leads to NULL Listener HQUIC
-      assert(l_ctx->Listener == NULL);
-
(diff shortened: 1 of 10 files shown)


c_src/quicer_listener.c
```diff
@@ -465,21 +463,140 @@ close_listener1(ErlNifEnv *env,
...
+}
+
+ERL_NIF_TERM
+start_listener3(ErlNifEnv *env,
+                __unused_parm__ int argc,
+                const ERL_NIF_TERM argv[])
+{
+  ERL_NIF_TERM listener_handle = argv[0];
+  ERL_NIF_TERM elisten_on = argv[1];
+  ERL_NIF_TERM options = argv[2];
```
thalesmg, line 537: Q: should we `assert(argc == 3)` before accessing those values?

thalesmg (approved):

c_src/quicer_listener.c
qzhuyan: could be done. I will add it in next PR. 
btw, the assertion in this project applies to TEST build only.

PR #210, opened 2023-08-30
refactor(test): move listener tests to quicer_listener_SUITE

qzhuyan:


test/quicer_SUITE.erl
```diff
@@ -366,222 +325,15 @@ tc_lib_re_registration(_Config) ->
...
-  PasswordCerts = [ {certfile, filename:join(DataDir, "server-password.pem")}
-                  , {keyfile,  filename:join(DataDir, "server-password.key")}
-                  , {password, ?SERVER_KEY_PASSWORD}
-                  ],
-  {ok, L} = quicer:listen(Port, default_listen_opts(PasswordCerts ++ Config)),
-  quicer:close_listener(L),
-  ok.
-
-tc_open_listener_with_wrong_cert_password(Config) ->
+tc_open_listener_inval_reg(Config) ->
```
thalesmg, line 328: Q: isn't this a listener test too?

thalesmg (approved):

test/quicer_SUITE.erl
qzhuyan: it should be but it is also testing a negative case when its dep (registration) is unavailable(closed). 

Actually, I moved it back here because due to the current limitation in msquic and quicer I could not close the global registration when I have lingering listener and connections which is created/used then forgotten in the suite. This tc must be run before any other test run otherwise it is deadlock.

PR #211, opened 2023-09-04
Support listener registration

### LINKED ISSUE #151: quicer:listen() twice
I cannot execute listen() twice in the same VM instance. 
Try

```

-module(abc).


-compile(export_all).
-compile(nowarn_export_all).


start() ->
    application:ensure_all_started(quicer),
    Port = 4567,
    LOptions = [ {cert, "./server.pem"}
               , {key,  "./server.key"}
               , {verify, none}
               , {active, false}
               , {alpn, ["sample"]}
               , {peer_bidi_stream_count, 64000}
               ],
    L = proc_lib:spawn_link(fun() ->
                          {ok, L} = quicer:listen(Port, LOptions),
                          listener(L)
                      end),
    L.

listener(_L) ->
    receive
        die ->
            ok
    end.
```

from the shell, in my book if the process dies, it should be possible to execute abc:start() again, bit it isn't
`listen()` then returns

```
{error,listener_start_error,{unknown_quic_status,91}}
```

To run: compile and then do:
```
Eshell V13.1.2  (abort with ^G)
1> abc:start().
<0.94.0>
2> a=p.
** exception error: no match of right hand side value p
3> abc:start().
<0.98.0>
4> =CRASH REPORT==== 8-Dec-2022::22:53:54.029762 ===

```

We earlier had a discussion on errno, regarding EMFILE, but 91 seems far fetched. It's "Protocol wrong type for socket"

qzhuyan:
followup #208

- Listen on specific registration with `quic_registration` listen opt. 
- Split large listen NIF functions to different stages.
- also fix #151 to stop listen when owner is dead.


c_src/quicer_config.c
```diff
@@ -2441,36 +2461,79 @@ get_str_from_map(ErlNifEnv *env,
...
+  if (!str_buffer)
+    {
+      str_buffer = (char *)malloc(len + 1);
+    }
+
+  if (enif_get_string(env, tmp_term, str_buffer, len + 1, ERL_NIF_LATIN1))
+    {
+      return str_buffer;
+    }
+
```
thalesmg, line 2514: Newbie question: if we allocated `str_buffer` but didn't return it above, should we free it here?

thalesmg (dismissed):

c_src/quicer_config.c
qzhuyan: yes. it should! thanks!

thalesmg (approved):

PR #212, opened 2023-09-04
Dev/william/msquic 2.2

qzhuyan:


test/quicer_snb_SUITE.erl
```diff
@@ -1701,16 +1701,20 @@ tc_conn_stop_notify_acceptor(Config) ->
...
                                                                        end, lists:seq(1, 100)),
                                                  {ok, Conn} = quicer:handshake(Conn),
-                                                 {error, closed} = quicer:accept_stream(Conn, []),
+                                                 case quicer:accept_stream(Conn, []) of
+                                                   {error, closed} -> ok;
+                                                   {ok, _Stream} -> ok
+                                                 end,
                                                  exit({normal, Acceptors})
                                              end),
+                 timer:sleep(1000),
```
qzhuyan, line 1717: should remove...

test/quicer_SUITE.erl
```diff
@@ -642,6 +642,9 @@ run_tc_conn_client_bad_cert(Config)->
         {error, stm_open_error, aborted} ->
           %% Depending on the timing, connection open could fail already.
           ok;
+        {error, stm_start_error, aborted} ->
+          %% Depending on the timing, connection open could fail already.
```
qzhuyan, line 646: s/open/start

CMakeLists.txt
```diff
@@ -21,12 +21,23 @@ else()
...
+
 if (DEFINED ENV{QUIC_ENABLE_LOGGING})
   set(QUIC_ENABLE_LOGGING $ENV{QUIC_ENABLE_LOGGING})
+  set(QUIC_LOGGING_TYPE lttng)
 else()
   set(QUIC_ENABLE_LOGGING "OFF")
 endif()
 
+if (DEFINED ENV{QUIC_LOGGING_TYPE})
+  set(QUIC_ENABLE_LOGGING "ON")
```
paulozulato, line 37: `QUIC_LOGGING_TYPE` can activate `QUIC_ENABLE_LOGGING` even though this one is not set. And vice versa. Therefore, shouldn't be better to merge this two variables in a single one? Like `QUIC_LOGGING` which could be `off`, `lttng`, etc (any value different from `off` would be considered the logging type and would imply that logging is on)?

thalesmg (approved):

CMakeLists.txt
qzhuyan: good idea but these are exposed by msquic and they are optional config,  I think it is just simple to keep them aligned with MsQuic.

paulozulato (approved):

ieQu1 (approved):

thalesmg (dismissed):

zmstone (approved):

PR #214, opened 2023-09-07
refactor:  QUIC datagram

qzhuyan:
Refactor dgram events, this is a breaking change.

Make dgram events follow others event format 4-element tuple. 

- dgram send state event
  Old code only report success of sending but not other progressing states (discard, cancelled, waiting ack ...)
  This PR makes it report any progress state so that application could get feedback quickly and react on it quickly. 

- dgram state event
  Old code only reports when it is enabled.
  This PR makes it also reports when it is disabled also with max_len for application layer feedback loop.

- dgram received data event
  use 4 elements tuples format. It looks like a stream data recv event but with connection handle. 

Need followups: 
1. Since this PR makes quicer emit more dgram related events  which should be masked by a send opt.
2. Behavior module need to be updated for handling these events and do callbacks.


src/quicer.erl
```diff
@@ -727,11 +727,20 @@ do_recv(Stream, Count, Buff) ->
         {error, dgram_send_error, atom_reason()}.
 send_dgram(Conn, Data) ->
   case quicer_nif:send_dgram(Conn, Data, _IsSync = 1) of
-    %% @todo make ref
+    %% @todo we need find tuned event mask
     {ok, _Len} = OK ->
       receive
-        {quic, send_dgram_completed, Conn} ->
-          OK
+        {quic, dgram_send_state, Conn, #{state := ?QUIC_DATAGRAM_SEND_SENT}} ->
```
qzhuyan, line 733: these are not nice, we could intro some dgram event mask for caller to skip uninterested events. also intro `async_send_dgram`.

thalesmg (approved):

PR #215, opened 2023-09-07
Dev/william/connection registrations

qzhuyan:
followup: #215,
 client can create connection with registration
 and some refactorings


c_src/quicer_connection.c
```diff
@@ -472,22 +494,47 @@ ServerConnectionCallback(HQUIC Connection,
...
-                 __unused_parm__ int argc,
-                 __unused_parm__ const ERL_NIF_TERM argv[])
+open_connectionX(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[])
 {
   QUIC_STATUS Status = QUIC_STATUS_SUCCESS;
   ERL_NIF_TERM eHandle;
   ERL_NIF_TERM res = ERROR_TUPLE_2(ATOM_ERROR_INTERNAL_ERROR);
+  QuicerRegistrationCTX *r_ctx = NULL;
+  HQUIC registration = NULL;
+  ERL_NIF_TERM options = argv[1];
```
thalesmg, line 510: Q: should we assert that `argc >= 2` before accessing this index?

thalesmg (approved):

c_src/quicer_connection.c
qzhuyan: yes, the current assertion only works with TEST macro and it will cause compiler warning, unused args without TEST macro.   

currently, they are safe since number of args are defined here: https://github.com/emqx/quic/blob/main/c_src/quicer_nif.c#L1480,
but I will fix them all together when I add unit test for C code.

PR #227, opened 2023-11-06
fixing release script

qzhuyan:


CMakeLists.txt
```diff
@@ -83,7 +83,7 @@ set(QUIC_TLS_SECRETS_SUPPORT "ON")
 
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/include/quicer_vsn.hrl.in
```
zmstone, line 85: the .hrl.in file should not be in `include` dir, maybe put it in another directory which is not going to be packed into a hex package.

zmstone:
q: why don't generate the hrl file and commit to git

qzhuyan:
> q: why don't generate the hrl file and commit to git

It is the same I think, or?

EMQX forked rebar3 copies the src to `_build` before `compile` so it is risky that rebar3 uses the old committed file. because the hrl generation is executed in a pre-compile hook we need to copy the generated file to `_build` anyway.  I had a similar issue while I was cheated to test old *.so file from the old build.

better idea?

thalesmg (approved):

PR #228, opened 2023-11-13
feat: cache stream id

### LINKED ISSUE #226: Can't get stream ID after receiving stream_closed message
This no longer works on main:

```
receive
         {quic, stream_closed, StreamRef, Flags} →
             {ok, StreamID} = quicer:get_stream_id(StreamRef),
             ...
end
```

Before I could get the StreamID, now I can't. It's fine if we can't use the StreamRef anymore, but I would expect to be able to retrieve metadata at least.

I guess it's not a big deal since I can get it from my state but it's surprising.

qzhuyan:
fix #226


essen:
I'll try this now, cheers.

PR #239, opened 2023-12-01
Fix/william/stream send after close

qzhuyan:


zmstone:
merge and release a new version ?

PR #243, opened 2023-12-06
feat: dead stream acceptor fallback to conn owner

qzhuyan:
without this change, peer will get stream send/recv aborted as we return an error (QUIC_STATUS_UNREACHABLE) to the callback if picked the owner who is dead.


qzhuyan:
thanks @thalesmg !

PR #251, opened 2023-12-20
Set cacertfile in tls ctx

### LINKED ISSUE #250: msquic 2.2 ca certificate file setting in SSL_CTX
https://github.com/microsoft/msquic/commit/a1762b1df44d9909cb995135c087d8c0d49d670d

qzhuyan:
fix #250 

This is new feature in msquic 2.2.

Before this change certificate (chain) validation was done in the callback (quicer) , now it is done in TLS.
- The old method is wrapped with `QUICER_USE_TRUSTED_STORE`
   We used to have trusted store cache in listener to speed up handshakes.
- more tweaks for macOS to make it work

### DIFF
--- a/.github/workflows/cover.yaml
+++ b/.github/workflows/cover.yaml
@@ -42,6 +42,21 @@ jobs:
         git-branch: ${{ github.ref_name }}
         git-commit: ${{ github.sha }}
 
+    - name: Run cover with feature switches
+      run: |
+        sudo apt-get install lcov
+        export QUICER_USE_TRUSTED_STORE=1
+        make cover
+
+    - name: Coveralls C
+      uses: coverallsapp/github-action@v2
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        flag-name: run-c-lcov-feature
+        parallel: true
+        git-branch: ${{ github.ref_name }}
+        git-commit: ${{ github.sha }}
+
     - name: Coveralls Erl
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/c_src/quicer_ctx.h
+++ b/c_src/quicer_ctx.h
@@ -97,9 +97,7 @@ typedef struct QuicerConnCTX
   ErlNifMutex *lock;
 #if defined(QUICER_USE_TRUSTED_STORE)
   X509_STORE *trusted;
-#else
-  char *cacertfile;
-#endif // !QUICER_USE_TRUSTED_STORE
+#endif // QUICER_USE_TRUSTED_STORE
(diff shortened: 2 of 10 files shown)


ieQu1 (dismissed):

c_src/quicer_listener.c
```diff
@@ -59,12 +59,13 @@ ServerListenerCallback(__unused_parm__ HQUIC Listener,
...
       CxPlatRefInitialize(&(c_ctx->ref_count));
 
+#if defined(QUICER_USE_TRUSTED_STORE)
       if (l_ctx->trusted_store)
         {
           X509_STORE_up_ref(l_ctx->trusted_store);
           c_ctx->trusted = l_ctx->trusted_store;
         }
-
+#endif // QUICER_ENABLE_DEBUG_MSG
```
thalesmg, line 68:
```suggestion
#endif // QUICER_USE_TRUSTED_STORE
```
?

c_src/quicer_ctx.h
```diff
@@ -93,11 +95,13 @@ typedef struct QuicerConnCTX
   ErlNifMonitor owner_mon;
   ErlNifEnv *env;
   ErlNifMutex *lock;
+#if defined(QUICER_USE_TRUSTED_STORE)
   X509_STORE *trusted;
-  // Connection handle closed flag
-  // false means the handle is invalid
+#endif // !QUICER_USE_TRUSTED_STORE
```
thalesmg, line 100: Note: the comment says `!QUICER_USE_TRUSTED_STORE`, but the condition is triggered when it's defined.

.github/workflows/cover.yaml
```diff
@@ -42,6 +42,21 @@ jobs:
...
+      run: |
+        sudo apt-get install lcov
+        export QUICER_USE_TRUSTED_STORE=1
+        make cover
+
+    - name: Coveralls C
+      uses: coverallsapp/github-action@v2
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        flag-name: run-c-lcov-feature
```
thalesmg, line 55: Just curiosity: is this flag defined somewhere in the codebase?  I didn't find it.  :see_no_evil:

thalesmg (approved):

.github/workflows/cover.yaml
qzhuyan: Just a name, let coverall knows it is different subtask or parallel job

c_src/quicer_ctx.h
qzhuyan: Good catch, I removed the #else ... part but leave this behind....

PR #291, opened 2024-06-25
feat(lb): user netdev ip as lb server fixed id

qzhuyan:


src/quicer_nif.erl
```diff
@@ -156,8 +156,15 @@ open_lib() ->
             {error, _} ->
                 priv_dir()
         end,
+    LBMode =
+        case application:get_env(quicer, lb_mode, 0) of
+            ?QUICER_LOAD_BALANCING_IFIP_AS_SERVER_ID ->
+                lb_server_id(ipv4, application:get_env(quicer, lb_dev, ?DEFAULT_LB_DEV));
```
keynslug, line 162: Nit. It feels pretty bold to hardcode "default" interface, since it is largely unpredictable how it looks on different systems. Since it is an application-level env, might just make it explicit by making `lb_mode` follow the typespec `0..3 | {if_ip, IfName}` or something like that. What do you think?

src/quicer_nif.erl
```diff
@@ -424,3 +431,20 @@ priv_dir() ->
...
+        {ok, IfList} = inet:getifaddrs(),
+        %% @NOTE Be aware of the order of the bytes in the address
+        lists:foldr(
+            fun(I, V) -> V bsl 8 bor I end,
+            0,
+            tuple_to_list(proplists:get_value(addr, proplists:get_value(DevName, IfList)))
+        )
+    catch
+        _:E ->
+            logger:error("Failed to set lb mode from ~s, fallback to disabled: ~p", [DevName, E]),
```
keynslug, line 448: Nit. I'd vote for more transparent error handling here, instead of having `{function_clause, ...}` errors with huge stacktraces in logs just because of `proplists:get_value(addr, undefined)` call.

qzhuyan: This is only called during module load, stacktrace won't be huge, and most likely shoudn't fail if config is correct. 

not only because of `proplists:get_value`, `inet:getifaddrs()` may also fail, not much we could do to handle the error here ,logging should be enough.

src/quicer_nif.erl
qzhuyan: good point. To make it simpler I think I just use the `IfName::list()` instead of `{if_ip, IfName}`.

src/quicer_nif.erl
thalesmg: Since it's during module load, maybe if it were to fail fast, the user would notice earlier that something is wrong/unexpected.

test/quicer_SUITE.erl
```diff
@@ -3081,6 +3082,25 @@ tc_stream_get_owner_remote(Config) ->
...
+        case os:type() of
+            {unix, darwin} -> "lo0";
+            _ -> "lo"
+        end,
+    application:set_env(quicer, lb_mode, NetDevName),
+    quicer:reg_close(),
+    quicer:close_lib(),
+    {ok, _} = quicer:open_lib(),
+    ?assertEqual(
+        {ok, 2},
```
thalesmg, line 3100: At first, this looks like it could be:

```suggestion
        {ok, ?QUIC_LOAD_BALANCING_SERVER_ID_FIXED},
```

But, if we set `lb_mode` to `lo`, shouldn't it be a different number, like `16777343`?

thalesmg (approved):

test/quicer_SUITE.erl
qzhuyan: 16777343 won't be a values supported by msquic, but quicer NIF use the magic to covert it to QUIC_LOAD_BALANCING_SERVER_ID_FIXED as the value of  `LoadBalancingMode` and at the same time set the `FixedServerID` to 16777343 which is a hint for custmized  LB.  

"lo" is just for testing. In real it should be the net dev that recv from the LB.

qzhuyan: I need to fix the release script as well, will update in that PR.

PR #297, opened 2024-09-20
build: remove dup dyn lib files for EMQX build

qzhuyan:
Downside is it won't be upgradable.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -68,6 +68,17 @@ download() {
     return $res
 }
 
+# rebar3 deref softlinks while packaging
+# so we dref at here to keep one dynlib file to avoid dup files in EMQX package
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -68,6 +68,17 @@ download() {
     return $res
 }
 
+# rebar3 deref softlinks while packaging
+# so we dref at here to keep one dynlib file to avoid dup files in EMQX package
+# - priv/libquicer_nif.so
+# - priv/libquicer_nif.so.1
+# - priv/libquicer_nif.so.504
```
id, line 75: what about `libmsquic `?

qzhuyan: it is already excluded from pkg.

id (approved):

PR #311, opened 2024-10-21
please debug emu

qzhuyan:
- extended asan scripts to run with more emu types to find bugs, (debug, valgraint...)
- fix for debug emu: 
   1. ensure all terms belongs to the same nif env
   2. ensure init atoms before first trace event (in snabbkaffe test)
   3. add `is_monitored` flag to avoid demonitor uninitiated monitor. 
- fix a issue that stream owner is not monitored by the nif resources .
- fix flaky common test
- fix falky proper test

however the debug emu is still not happy with snabbkaffe trace event when calling whereis(snabbkaffe_collector). 
This only effect test with snabbkaffe,  May fire an OTP issue for that.


tools/run/bin/debug-check
```diff
@@ -0,0 +1,53 @@
...
+            for tc in ${AllTCs};
+            do
+                echo "running tc $tc";
+                escript "$REBAR3" do ct --suite=test/quicer_SUITE --case="$tc";
+            done
+            ;;
+        proper)
+            escript "$REBAR3" as test proper
+            ;;
+    esac
```
thalesmg, line 50: Micro nit (otherwise might fail silently in CI due to typos):

```suggestion
            ;;
        *)
            echo "Invalid option: $1"
            exit 1
            ;;
    esac
```

thalesmg: (same for valgrind-check)

c_src/quicer_connection.c
```diff
@@ -841,7 +841,14 @@ async_connect3(ErlNifEnv *env,
 
   // Monitor owner before start, so we don't need to race with callbacks
   // after start the connection
-  enif_monitor_process(NULL, c_ctx, &c_ctx->owner->Pid, &c_ctx->owner_mon);
+  //
+  if (!c_ctx->is_monitored
```
thalesmg, line 845: Just noting: l. 1179 is slightly different: `(!c_ctx->is_monitored) && 0 == ...` versus `!c_ctx->is_monitored && 0 == ...`, in case C associativity rules make a difference here.

test/example_client_connection.erl
```diff
@@ -152,17 +153,19 @@ peer_needs_streams(C, bidi_streams, S) ->
...
+    case lists:keyfind(Stream, 2, Streams) of
+        {Reason, Stream} when
+            Reason =:= owner_down orelse
+                Reason =:= closed orelse
+                Reason =:= start_error
+        ->
+            _ = quicer:async_shutdown_stream(Stream),
+            {ok, S#{streams := lists:keydelete(Stream, 2, Streams)}};
+        {OwnerPid, Stream} when is_pid(OwnerPid) ->
+            {error, {fixme, bug_handoff_fail}};
```
thalesmg, line 167: Supposed to fix in this PR?  Otherwise might be useful as a reference to add comment about this fixme.

thalesmg (approved):

qzhuyan:
Thanks for review, will address the comments in a follow up

PR #315, opened 2024-10-24
test: use global_snab_kc_pid

qzhuyan:
fix lock_order_violation when run with debug emu type
````
    at beam/erl_lock_check.c:1225
    at beam/erl_threads.h:1991
````

debug emu doesn't like to call enif_whereis_pid in resource down callback.

### DIFF
--- a/test/quicer_snb_SUITE.erl
+++ b/test/quicer_snb_SUITE.erl
@@ -66,6 +66,17 @@
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 -include("quicer.hrl").
 
+-define(my_check_trace(BUCKET, RUN, CHECK),
+    ?check_trace(
+        BUCKET,
+        begin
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/quicer_snb_SUITE.erl [hunk 1]
```diff
@@ -66,6 +66,17 @@
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 -include("quicer.hrl").
 
+-define(my_check_trace(BUCKET, RUN, CHECK),
+    ?check_trace(
+        BUCKET,
```
ieQu1, line 71: Why is this argument called BUCKET?

ieQu1 (approved):

test/quicer_snb_SUITE.erl
qzhuyan: copied from https://github.com/kafka4beam/snabbkaffe/blob/master/include/test_macros.hrl#L108

PR #331, opened 2024-12-18
Fix MacOS 15 issue with M3

danicuki:


qzhuyan:
I need to hold this PR as I want to understand why cmake could not detect the arch correctly in the fist place.

qzhuyan:
sorry for parking this PR for too long, I will try to reproudce it when I have a mix setup on macOS.

qzhuyan:
solved in #414

PR #338, opened 2025-01-14
ci: switch workflows to recent Erlang/OTP 27

keynslug:
Another notable change to the release workflow is an ability to re-release existing tags through the _workflow dispatch_ trigger. This could be useful to populate existing releases with binaries built for recent Erlang/OTP versions (in this case Erlang/OTP 27).

See individual commits for details.

### DIFF
--- a/test/quicer_test_lib.erl
+++ b/test/quicer_test_lib.erl
@@ -402,7 +402,13 @@ report_active_connections() ->
     report_active_connections(fun ct:comment/2).
 report_active_connections(LogFun) ->
     erlang:garbage_collect(),
-    {ok, Cnts} = quicer:perf_counters(),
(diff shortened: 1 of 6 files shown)


test/quicer_test_lib.erl
```diff
@@ -402,7 +402,13 @@ report_active_connections() ->
     report_active_connections(fun ct:comment/2).
 report_active_connections(LogFun) ->
     erlang:garbage_collect(),
-    {ok, Cnts} = quicer:perf_counters(),
+    case quicer:perf_counters() of
+        {ok, Cnts} ->
+            ok;
+        {error, not_found} ->
```
qzhuyan, line 408: Need follow up on this.

PR #342, opened 2025-01-28
Guard glibc-specific functions using __GLIBC__ define

Jolmberg:
Hi! I've been compiling and running quicer on alpine, but recent versions no longer compile. I'm a little bit out of my depth here but it seems to be because alpine uses musl instead of glibc and is missing the malloc_trim and malloc_stats functions. These are already protected by the __linux__ define so I'm wondering if you would consider changing that to __GLIBC__ instead.

### DIFF
--- a/c_src/quicer_nif.c
+++ b/c_src/quicer_nif.c
@@ -17,11 +17,12 @@ limitations under the License.
 #include "quicer_nif.h"
 
 #include <dlfcn.h>
+#include <features.h>
 
 #include "quicer_listener.h"
 #include "quicer_vsn.h"
 
-#if defined(__linux__)
+#if defined(__GLIBC__)
 #include <malloc.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


qzhuyan:
@Jolmberg  thanks for the PR.

Could you fix the formatting and update your branch? 

```
diff --git a/c_src/quicer_nif.c b/c_src/quicer_nif.c
index 0b5b658..6b45bd7 100644
--- a/c_src/quicer_nif.c
+++ b/c_src/quicer_nif.c
@@ -16,8 +16,8 @@ limitations under the License.
 
 #include "quicer_nif.h"
 
-#include <features.h>
 #include <dlfcn.h>
+#include <features.h>
 
 #include "quicer_listener.h"
 #include "quicer_vsn.h"
```

Jolmberg:
Thank you!

PR #361, opened 2025-07-28
ci: update release workflow

id:
Use latest builder version, add missing operating systems


qzhuyan:
I think github just dont give us enough credit to build so many varients.

PR #368, opened 2025-08-15
feat(tls): custom verify certs

### LINKED ISSUE #332: Support for custom cert validation
Hey I needed a way to inspect a cert before accepting a connection and implemented it on [this fork](https://github.com/emqx/quic/compare/main...JamBrains:quic:main?expand=1) by adding four things:
- Always include `QUIC_CREDENTIAL_FLAG_INDICATE_CERTIFICATE_RECEIVED` and `QUIC_CREDENTIAL_FLAG_REQUIRE_CLIENT_AUTHENTICATION` into the `CredConfig` returned by `parse_verify_options`.
- Send an event `ATOM_PEER_CERT_RECEIVED` to the connection owner in `handle_connection_event_peer_certificate_received` to notify them of a received cert.
- Added function `complete_cert_validation2(conn, bool)` to accept / reject an inbound cert.
- Changed `handle_connection_event_peer_certificate_received` to return `QUIC_STATUS_PENDING` instead of success.

This makes it possible to asynchronously inspect a cert and accept or reject it before the connection is established (possibly also interesting for @danicuki). 

Do you think this is generally the right direction? The `parse_verify_options` would probably need to have some arguments changed to make it possible to indicate the custom cert checking instead of always enabling it. And I would do some other cleanups.

qzhuyan:
fix #332


docs/internals.org
```diff
@@ -275,7 +275,7 @@ SYNC call in non-callback-context
 | QUIC_CONNECTION_EVENT_DATAGRAM_SEND_STATE_CHANGED     | {quic, send_dgram_completed, connection_handle()}                         |
 | QUIC_CONNECTION_EVENT_RESUMED                         | {quic, connection_resumed, connection_handle, false \vert binary()}       |
 | QUIC_CONNECTION_EVENT_RESUMPTION_TICKET_RECEIVED      | {quic, nst_received, connection_handle(), Ticket::binary()}               |
-| QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED       | @TODO                                                                      |
+| QUIC_CONNECTION_EVENT_PEER_CERTIFICATE_RECEIVED       | {quic, peer_cert_received, connection_handle(), Cert::binary()}            |
```
qzhuyan, line 278: Prefer to pass the cert chain instead.

Issue #335, opened 2025-01-05
Support in-memory Cert and Key passing


ggwpez:
The credential opts require `certfile`, `keyfile` and possibly `cacertfile`. This is a bit impractical since i generate lots of Cert-Key pairs locally for testing (thousands).  

The [commit](https://github.com/JamBrains/quic/commit/7b4e5aa00970fd6b57417fe33010689799752c21) here adds a `certkeyasn1` option that can be used to pass a PKCS12-ASN.1 encoded blob, according to [this](https://github.com/microsoft/msquic/blob/main/docs/api/QUIC_CREDENTIAL_CONFIG.md#type).  
I will clean it up a bit when I get the time, but it seems to work so far. It is currently missing password support.

qzhuyan:
cool, just open the PR and we could co-auth.

qzhuyan:
any plan to open a PR?

ggwpez:
Hey sorry for the long reply time, I am working on the tests for this now. Will submit the MR by EOD

PR #369, opened 2025-08-16
Support in-memory Cert and Key options

ggwpez:
Closes https://github.com/emqx/quic/issues/335

Changes:
- Add `certkeyasn1` to the listener options. This allows to pass a PKCS12 ANS1 encoded binary that contains a certificate and a key directly to the listener instead of having to pass file paths. This is useful when spawning 100s-1000s of listeners with different certs and keys.
- It currently ignores the `password` option and returns an error if it is specified.
- Either `certkeyasn1` OR (`certfile` and `keyfile`) can be set. Any other combination results in an option parsing error.


c_src/quicer_tls.c
```diff
@@ -16,10 +16,73 @@ limitations under the License.
...
+    }
+
+    CertInMem = (QUIC_CERTIFICATE_PKCS12 *)CXPLAT_ALLOC_NONPAGED(
+            sizeof(QUIC_CERTIFICATE_PKCS12), QUICER_CERTIFICATE_PKCS12);
+    if (!CertInMem)
+      {
+        goto error;
+      }
+
+    char *allocated_asn1 = (char *)CXPLAT_ALLOC_NONPAGED(asn1_bin.size, QUIC_POOL_TEST);
```
ggwpez, line 60: Not sure here

test/quicer_SUITE.erl
```diff
@@ -3668,7 +3668,7 @@ default_listen_opts_client_cert(Config) ->
     [
         {cacertfile, filename:join(DataDir, "ca.pem")},
         {verify, peer}
-        | tl(default_listen_opts(Config))
+        | proplists:delete(verify, default_listen_opts(Config))
```
ggwpez, line 3671: This was assuming that `verify` is the first entry.

ggwpez:
Hey @qzhuyan, sorry for being so slow here and thanks for your patience!  
Yes you can always push my branches, i dont mind.

c_src/quicer_tls.c
```diff
@@ -16,10 +16,73 @@ limitations under the License.
...
+    if (!allocated_asn1)
+      {
+        goto error;
+      }
+    CxPlatCopyMemory(allocated_asn1, asn1_bin.data, asn1_bin.size);
+
+    // Tie it all together:
+    CertInMem->Asn1BlobLength = asn1_bin.size;
+    CertInMem->Asn1Blob = (uint8_t const*)allocated_asn1;
```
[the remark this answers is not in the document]
ggwpez: Done and set `CertificatePkcs12` to NULL as well.

qzhuyan:
Tests passed on my Mac 15 m2 locally with openssl3 or openssl. 

must be something special in github macos runner.

qzhuyan (approved):

PR #376, opened 2025-09-01
feat: runtime envvar kill switch QUICER_NO_NIF

qzhuyan:
fix [EEC-951](https://emqx.atlassian.net/browse/EEC-951)

By setting QUICER_NO_NIF=1,
NIF module (quicer_nif) will be loaded successfully, and quicer
application could be started but NONE of the functions can work.

This is used in some deployment that the quicer functions are not
required/desired but quicer app is embedded in release that requires
it to start successfully.

### DIFF
--- a/src/quicer_nif.erl
+++ b/src/quicer_nif.erl
@@ -119,7 +119,7 @@
 -include("quicer_types.hrl").
 -include("quicer_vsn.hrl").
 
--define(NONIF_SWITCH, "QUICER_NO_NIF").
+-define(NONIF_SWITCH, "QUICER_SKIP_NIF_LOAD").
 
 -hank([
     {unnecessary_function_arguments, [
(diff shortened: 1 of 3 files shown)


src/quicer_nif.erl
```diff
@@ -133,6 +142,19 @@ init() ->
     init(ABIVsn).
 
 init(ABIVsn) ->
+    case os:getenv(?NONIF_SWITCH) of
+        "1" ->
+            io:format(
+                "~n~nWARN: Detected envvar ~s=1, ~p module is loaded but NOT functional. This is not encouraged and take your own risk!~n~n",
+                [?NONIF_SWITCH, ?MODULE]
+            ),
```
zmstone, line 150:
```suggestion
            io:format(
                "~n~nWARN: Detected env ~s=1, QUIC module is loaded but will NOT be functional.~n",
                [?NONIF_SWITCH]
            ),
```

zmstone (approved):

PR #403, opened 2026-03-14
Fix representation of option types

jcpetruzza:
In `quicer_types.hrl`, option types are represented as maps. There are however a couple of issues that show up when trying to use quicer as a dependency in project that runs dialyzer or a type-cheker like eqwalizer:

1. For clarity and maintainability, there are types defined like this:
```
-type listen_opts() :: listen_security_opts() | quic_settings().
```
Here the idea is that the `listen_opts()` is the map-union of the `listen_security_opts()` and the `quic_settings()`. However, in Erlang's language of types, this instead should be interpreted as: a value of type `listen_opts()` is EITHER a `listen_security_opts()` or a `quic_settings()`. So both dialyzer and eqwalizer may flag code that tries to call a function expecting a `listen_opts()`, etc.

To workaround this, we can rely on macros to define the map fields that are common to various option types.

- Keys like `key`, `keyfile`, `cert` and `certfile` are all marked as "mandatory", however, callers are expected to pass either `cert` or a `certfile` but not both. With the current definition, if one tries to pass only a `certfile`, a type-checker will complaint that the `cert` key is missing, etc.

Here, it is necessary to drop the mandatory annotation.

- For user payload, currently `_ => _` is used, which is short for `term() => term()`. However, a type like `_ => dynamic()` would be preferable here, as it better represents a type that is "fixed (by the user),, but unknown".

### DIFF
--- a/include/quicer_types.hrl
+++ b/include/quicer_types.hrl
@@ -20,7 +20,7 @@
 -include("quicer.hrl").
 
 %% https://github.com/AdRoll/rebar3_hank/issues/155
--hank([{unused_macros, [{"MASK", 1}]}]).
+-hank([{unused_macros, [{"MASK", 1}, {"TLS_OPTS", 0}, {"QUIC_SETTINGS_OPTS", 0}]}]).
 
 -define(BIT(Bits), (1 bsl (Bits))).
 -define(MASK(Bits), (?BIT(Bits) - 1)).
--- a/include/quicer_types.hrl
+++ b/include/quicer_types.hrl
@@ -27,8 +27,59 @@
 
 -export_type([handle/0]).
 
+-define(TLS_OPTS,
+    cert => file:filename(),
+    certfile => file:filename(),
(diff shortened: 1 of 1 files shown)


include/quicer_types.hrl
```diff
@@ -20,15 +20,66 @@
 -include("quicer.hrl").
 
 %% https://github.com/AdRoll/rebar3_hank/issues/155
--hank([{unused_macros, [{"MASK", 1}]}]).
+-hank([{unused_macros, [{"MASK", 1}, {"TLS_OPTS", 0}, {"QUIC_SETTINGS_OPTS", 0}]}]).
```
qzhuyan:
```suggestion
-hank([{unused_macros, [{"MASK", 1}, {"TLS_OPTS", none}, {"QUIC_SETTINGS_OPTS", none}]}]).
```

qzhuyan (approved):

Issue #377, opened 2025-09-12
build: github runner macOS build failed with Cmake 4


qzhuyan:
found issue in #368 

with github macOS 15 runner image:  '2.328.0', CMake 4.1.1

may relates to https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

```
- The C compiler identification is AppleClang 17.0.0.17000013
-- The CXX compiler identification is AppleClang 17.0.0.17000013
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Clang version >= 16.0 detected, enabling '-Wno-invalid-unevaluated-string'.
-- CMAKE Version: 4.1.1
-- Source Dir: /Users/runner/work/quic/quic/msquic
-- Host System name: Darwin
-- Setting policy 0091
-- System name: Darwin
-- System version: 24.5.0
-- Platform version: 
-- Build type: RelWithDebInfo
-- QUIC Platform: darwin
-- Version Build ID: 0
-- Version Suffix: -private
-- Looking for _SC_PHYS_PAGES
-- Looking for _SC_PHYS_PAGES - found
-- Looking for sysconf
-- Looking for sysconf - found
-- Looking for sysctl
-- Looking for sysctl - found
-- Enabling OpenSsl configuration tests
-- Enabling CA file tests
-- QUIC_ENABLE_LOGGING is false. Disabling logging
-- Disabling (client) shared port support
-- Configuring for OpenSSL 1.1
-- Setting openssldir to /opt/homebrew/etc/openssl@1.1
ERRORWTF darwin 
[...]

Issue #329, opened 2024-12-17
Error trying to compile quicer on MacOS Apple Silicon


danicuki:
Hi, I am trying to use this library for my Elixir project and I get this error when I try to compile. How to solve this?

`mix.exs dep: {:quicer, github: "emqx/quic"}`

Apple M3 Max - MacOS 15.1.1 (24B91)

```
mix --version
Erlang/OTP 26 [erts-14.2.5] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]

Mix 1.17.2 (compiled with Erlang/OTP 26)
```

```
 ~/dev/jamixir (networking)₿ mix deps.compile quicer --force
./build.sh 'v2.3.8'
allexport      	off
braceexpand    	on
emacs          	off
errexit        	on
errtrace       	off
functrace      	off
hashall        	on
histexpand     	off
history        	off
ignoreeof      	off
interactive-comments	on
keyword        	off
monitor        	off
noclobber      	off
noexec         	off
noglob         	off
nolog          	off
notify         	off
nounset        	on
onecmd         	off
physical       	off
pipefail       	off
posix          	on
privileged     	off
verbose        	off
vi             	off
xtrace         	off
Patching Msquic 2.3.8
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1539  100  1539    0     0   4444      0 --:--:-- --:--:-- --:--:--  4447
2 out of 2 hunks failed while patching 'src/inc/msquic_posix.h'
Skip patching patches/12edf3725475d4a99e5598df3289bace47b8f56e.patch, already applied
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   930  100   930    0     0   3243      0 --:--:-- --:--:-- --:--:--  3251
1 out of 1 hunks failed while patching 'src/core/binding.c'
Skip patching patches/e0201eb4e007e7524aef007be67f2281f949f102.patch, already applied
-- CMAKE Version: 3.28.3
-- Source Dir: /Users/danicuki/dev/jamixir/deps/quicer/msquic
-- Host System name: Darwin
-- Setting policy 0091
-- System name: Darwin
-- System version: 24.1.0
-- Platform version:
-- Build type: RelWithDebInfo
-- QUIC Platform: darwin
-- Version Build ID: 0
-- Version Suffix: -private
[...]

PR #414, opened 2026-05-06
build: detect macOS architecture for CMake

### LINKED ISSUE #413: macOS build should set CMAKE_OSX_ARCHITECTURES explicitly
## Summary

On macOS, `build.sh` does not set `CMAKE_OSX_ARCHITECTURES`. With newer CMake/macOS combinations this can leave `CMAKE_OSX_ARCHITECTURES` empty during the msquic/OpenSSL configure step.

msquic's OpenSSL/QuicTLS CMake logic expects `CMAKE_OSX_ARCHITECTURES` to be `arm64` or `x86_64` on Darwin. If it is empty, it falls through to the fallback `config` path (printing `ERRORWTF darwin` in older msquic versions), which can produce architecture mismatches or bad OpenSSL compiler flags.

This appears related to #329 and #377, but the remaining issue is that the architecture is still implicit.

## Observed failure

In a Mix project pinned to `{:quicer, git: "https://github.com/emqx/quic.git", tag: "0.2.4"}`, on Apple Silicon with CMake 4.3.1:

```text
-- Configuring for OpenSSL 1.1
ERRORWTF darwin
-- Configuring OpenSSL: .../openssl/config ... ;-isysroot
clang: warning: no such sysroot directory: '-DL_ENDIAN' [-Wmissing-sysroot]
fatal error: 'stdlib.h' file not found
```

The generated CMake cache had:

```text
CMAKE_OSX_ARCHITECTURES:STRING=
CMAKE_OSX_SYSROOT:STRING=
```

The current `main` branch already sets `SDKROOT`, which addresses the empty sysroot part from #377, but `CMAKE_OSX_ARCHITECTURES` is still unset unless the caller provides it.

## Expected behavior

On macOS, `build.sh` should set `CMAKE_OSX_ARCHITECTURES` when the caller has not provided it. For a NIF, the safest default is the architecture of the running Erlang VM, not necessarily the shell architecture, because Rosetta can make `uname -m` misleading.

It should still preserve an explicit user-provided `CMAKE_OSX_ARCHITECTURES` for cross/custom builds.

## Verification

With a patch that derives the default from `erlang:system_info(system_architecture)`:

```text
-- Configuring OpenSSL: ARCHFLAGS="-arch arm64";.../quictls/Configure;darwin64-arm64-cc ... ;-isysroot;/Applications/Xcode.app/.../MacOSX26.4.sdk
CMAKE_OSX_ARCHITECTURES:STRING=arm64
CMAKE_OSX_SYSROOT:PATH=/Applications/Xcode.app/.../MacOSX26.4.sdk
```

Verified locally on Apple Silicon/macOS with:

```sh
bash -n build.sh
rm -rf c_build priv _build msquic && mkdir -p priv && ./build.sh v2.5.5
~/.mix/elixir/1-19-otp-28/rebar3 compile
```

dmorn:
Fixes #413.
Related to #329 and #377.

## What changed

`build.sh` now sets `CMAKE_OSX_ARCHITECTURES` on macOS when the caller has not already provided it.

The default is derived from the running Erlang VM via `erlang:system_info(system_architecture)`, with a `uname -m` fallback. This keeps the NIF architecture aligned with the VM architecture and avoids Rosetta shell/host mismatches.

The existing `SDKROOT` setup now also preserves a caller-provided `SDKROOT` instead of unconditionally overwriting it.

## Why

msquic's OpenSSL/QuicTLS Darwin configure path expects `CMAKE_OSX_ARCHITECTURES` to be `arm64` or `x86_64`. If CMake leaves it empty, the build can fall through to the generic OpenSSL `config` path (`ERRORWTF darwin` in older msquic), causing architecture mismatch and/or bad `-isysroot` compiler flags.

## Verified

On Apple Silicon/macOS:

```sh
bash -n build.sh
rm -rf c_build priv _build msquic && mkdir -p priv && ./build.sh v2.5.5
~/.mix/elixir/1-19-otp-28/rebar3 compile
```

Confirmed generated cache/configure output includes:

```text
CMAKE_OSX_ARCHITECTURES:STRING=arm64
CMAKE_OSX_SYSROOT:PATH=/Applications/Xcode.app/.../MacOSX26.4.sdk
-- Configuring OpenSSL: ARCHFLAGS="-arch arm64";...;darwin64-arm64-cc ...
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -6,6 +6,40 @@ MSQUIC_VERSION="$1"
 TARGET_SO='priv/libquicer_nif.so'
 PKGNAME="$(./pkgname.sh)"
 
+detect_macos_arch() {
+    local arch=""
+
+    # The NIF must match the architecture of the running Erlang VM. Prefer
+    # Erlang's own target triplet so Rosetta shells don't accidentally choose
+    # the host/shell architecture instead of the VM architecture.
+    if command -v erl >/dev/null 2>&1; then
+        arch="$(erl -noshell -eval 'io:format("~s", [erlang:system_info(system_architecture)]), halt().' 2>/dev/null | cut -d- -f1 || true)"
+    fi
+
+    case "$arch" in
+        aarch64 | arm64)
+            echo arm64
+            ;;
+        x86_64 | amd64)
+            echo x86_64
+            ;;
+        *)
+            arch="$(uname -m)"
+            case "$arch" in
+                aarch64 | arm64)
+                    echo arm64
+                    ;;
+                x86_64 | amd64)
+                    echo x86_64
+                    ;;
+                *)
+                    echo "$arch"
+                    ;;
+            esac
+            ;;
+    esac
+}
+
 build() {
     # default: 4 concurrent jobs
     JOBS=4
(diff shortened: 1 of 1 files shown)


dmorn:
Kinda duplicate of https://github.com/emqx/quic/pull/331

qzhuyan (approved):

PR #418, opened 2026-06-19
Export quicer_nif:open_lib/1

zhongwencool:
## Summary

- export `quicer_nif:open_lib/1` from the Erlang module
- keep the Erlang export table aligned with the C NIF function table

## Root cause

`c_src/quicer_nif.c` registers `open_lib/1`, but `src/quicer_nif.erl` only exported `open_lib/0`. In embedded/on-load flows this can make the NIF load fail because the NIF table contains a function that is not exported by the Erlang module.

## Validation

- Ran a static check that compares NIF functions registered in `c_src/quicer_nif.c` with exports declared in `src/quicer_nif.erl`; missing count is now `0`.
- Ran `rebar3 compile` successfully, including the msquic/quictls native build and quicer compilation.

### DIFF
--- a/src/quicer_nif.erl
+++ b/src/quicer_nif.erl
@@ -16,6 +16,7 @@
 -module(quicer_nif).
 -export([
     open_lib/0,
+    open_lib/1,
     close_lib/0,
     reg_open/0,
     reg_open/1,


qzhuyan:
Do you have error messages, pls? I don't think it is required to export all the nif functions in 'NIF module'

zhongwencool:
[nif-export-repro.zip](https://github.com/user-attachments/files/29169748/nif-export-repro.zip)

```bash 
cat .tool-versions
elixir 1.20.1-otp-29
erlang 29.0
```
I enable the inline compile options globally:

`ERL_COMPILER_OPTIONS='[inline,inline_list_funcs,{inline_size,64}]'`

```
nif-export-repro >  ./run.sh
bad_nif_export_repro.erl:6:5: Warning: inlining is enabled - local calls to NIFs may call their Erlang implementation instead
%    6|     erlang:load_nif(os:getenv("NIF_EXPORT_REPRO_NIF"), 0).
%     |     ^

good_nif_export_repro.erl:6:5: Warning: inlining is enabled - local calls to NIFs may call their Erlang implementation instead
%    6|     erlang:load_nif(os:getenv("NIF_EXPORT_REPRO_NIF"), 0).
%     |     ^

== bad module: foo/1 is local and gets inlined away ==
[{exports,[{foo,0},{load,0},{module_info,0},{module_info,1}]},{locals,[]}]
erl status: 0
{error,on_load_failure}
=WARNING REPORT==== 21-Jun-2026::09:46:11.246596 ===
The on_load function for module bad_nif_export_repro returned:
{error,{bad_lib,"Function not found bad_nif_export_repro:foo/1"}}


== good module: foo/1 is exported and survives for NIF binding ==
[{exports,[{foo,0},{foo,1},{load,0},{module_info,0},{module_info,1}]},
 {locals,[]}]
erl status: 0
{module,good_nif_export_repro}

PASS: NIF registered foo/1 fails when Erlang does not export it, and loads when exported.
```


> Do you have error messages, pls? I don't think it is required to export all the nif functions in 'NIF module'

qzhuyan (approved):

qzhuyan:
@zhongwencool thanks for clarifying!

PR #420, opened 2026-06-26
test upgrade from 0.2.5

qzhuyan:
- add hot upgrades tests from old 0.2.5
- clarify hot upgrade only support from patched version due to msquic
    hot upgrade without any listener, conns are fine but still not suggested.
- ci pipeline maintenance 

caught new issue: #421 #422


zmstone (approved):

qzhuyan:
note: cover test is failed due to known issue.  #421

PR #430, opened 2026-07-27
build: demote gcc maybe-uninitialized to warning (main-0.2)

zmstone:
## What

When the C compiler is GCC, add `-Wno-error=maybe-uninitialized` (directory-wide, before `add_subdirectory(msquic)`), so the one diagnostic stays visible as a warning instead of failing the build via msquic's hardcoded `-Werror`.

## Why

GCC 14.3.1 (the el10 / RHEL 10 toolchain used by emqx-builder 6.1-8) reports a `maybe-uninitialized` false positive in msquic v2.3.8:

```
src/core/packet.h:342:71: error: 'PacketNumber' may be used uninitialized [-Werror=maybe-uninitialized]
    inlined from 'QuicPacketEncodeShortHeaderV1' at src/core/packet.h:578:5
```

The same code is unchanged in msquic v2.5.0, so there is no upstream fix to backport; GCC 14.2 (debian13) does not report it — it's specific to GCC 14.3's inliner. This currently breaks building quicer from source on el10, where no prebuilt release asset exists (first hit in emqx/emqx el10 package builds: https://github.com/emqx/emqx/actions/runs/30258610690).

Scoped to `CMAKE_C_COMPILER_ID STREQUAL "GNU"` so clang/AppleClang never see the flag (clang has no `maybe-uninitialized` group). Follows the existing pattern of the Clang >= 16 `-Wno-invalid-unevaluated-string` block.

Once this lands in a 0.2.x tag, the temporary rebar-override workaround in emqx/emqx#18126 can be reverted in favor of a plain tag bump.

Verified locally: clean `make build-nif` from-source build; msquic core target `flags.make` shows `-Wno-error=maybe-uninitialized` alongside msquic's `-Werror`, and `-Wno-error=X` takes precedence for that diagnostic regardless of order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CyC5e51AUx2xww3LgDkHXK

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -106,6 +106,45 @@ jobs:
             openssl: sys
           - os: el7
             openssl: openssl3
+        ## el10 builder images only exist since emqx-builder 6.1 and
+        ## (as of now) only for otp 27, hence include instead of matrix os
+        include:
+          - erlang:
+              otp: 27.3.4.2-8
+              builder: 6.1-8:1.18.3-27.3.4.2-8
+            os: el10
+            arch: amd64
+            openssl: sys
+          - erlang:
+              otp: 27.3.4.2-8
+              builder: 6.1-8:1.18.3-27.3.4.2-8
+            os: el10
+            arch: arm64
+            openssl: sys
+          - erlang:
+              otp: 27.3.4.2-8
+              builder: 6.1-8:1.18.3-27.3.4.2-8
+            os: el10
+            arch: amd64
+            openssl: openssl3
+          - erlang:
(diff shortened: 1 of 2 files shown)


zmstone:
Added a second commit: `ci: release el10 prebuilt packages` — extends the release matrix with el10 (via `include`, otp 27 / emqx-builder 6.1-8 only, since el10 builder images don't exist for otp 26/28). With prebuilt el10 assets, emqx el10 package builds will download the NIF instead of compiling msquic from source.

qzhuyan (approved):

PR #431, opened 2026-07-27
build: demote gcc maybe-uninitialized to warning

zmstone:
Port of #430 (see it for full context) to `main`.

GCC 14.3.1 (el10 / RHEL 10 toolchain) reports a `maybe-uninitialized` false positive in msquic's `src/core/packet.h` (`QuicPktNumEncode`), fatal under msquic's hardcoded `-Werror` when building from source. Add `-Wno-error=maybe-uninitialized` for GCC only (clang has no such warning group), keeping the diagnostic visible as a warning.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CyC5e51AUx2xww3LgDkHXK

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -74,11 +74,11 @@ jobs:
       matrix:
         erlang:
           - otp: 26.2.5.14-1
-            builder: 5.6-2:1.15.7-26.2.5.14-1
-          - otp: 27.3.4.2-3
-            builder: 5.6-2:1.18.3-27.3.4.2-3
-          - otp: 28.2-2
-            builder: 6.0-9:1.19.1-28.2-2
+            builder: 6.1-8:1.15.7-26.2.5.14-1
+          - otp: 27.3.4.2-8
+            builder: 6.1-8:1.18.3-27.3.4.2-8
+          - otp: 28.4.1-4
+            builder: 6.1-8:1.19.1-28.4.1-4
 
         openssl:
           - quictls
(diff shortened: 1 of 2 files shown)


zmstone:
Same as #430: second commit extends the release matrix with el10 (quictls variant, otp 27 / emqx-builder 6.1-8, amd64+arm64).

zmstone:
Reworked per review direction: instead of el10-only include entries, all erlang rows now use emqx-builder **6.1-8**, which provides images for every otp line and every distro including el10 — so `el10` is simply added to the `os` list. This also bumps otp 27.3.4.2-3 → 27.3.4.2-8 and 28.2-2 → 28.4.1-4 (asset names carry only the otp major, so consumers are unaffected). Verified the exact image tags exist on ghcr for el10/el7/amzn2/ubuntu20.04 across the new rows.

qzhuyan:
since this is main branch, maybe consider patch msquic instead.

qzhuyan:
Strange, I cannot reproduce the issue with image: ghcr.io/emqx/emqx-builder/6.1-8:1.18.3-27.3.4.2-8-el10 with `main` branch

```
docker run --rm     -v "$PWD:/main"     ghcr.io/emqx/emqx-builder/6.1-8:1.18.3-27.3.4.2-8-el10     sh -lc '
      set -e
      git config --global --add safe.directory /main
      git config --global --add safe.directory /main/msquic
      cd main
 make build-nif'
```

qzhuyan:
testing rel build here: https://github.com/qzhuyan/quic/actions/runs/30990014845

PR #432, opened 2026-08-04
Fix missing entries in statistics_v2 proplist

jcpetruzza:
`enif_make_list()` element count (36) was smaller than the 39 tuples passed, silently dropping `send_congestion_window`, `dest_cid_update_count`, and `send_ecn_congestion_count` from the resulting proplist.


qzhuyan (approved):

qzhuyan:
@jcpetruzza thanks for the PR.

PR #437, opened 2026-08-12
fix: clear is_recv_pending after passive->active re-arm completes the pending recv

jbevemyr:
### The bug

`set_stream_opt` handles the passive->active transition by re-enabling receives and, when `is_recv_pending` is set, calling `StreamReceiveComplete(0)` so held data is re-indicated. But `is_recv_pending` is never cleared afterwards. Once a stream has answered a single receive with `QUIC_STATUS_PENDING` in passive mode (no `recv()` caller waiting), the flag stays true forever, and every later passive->active transition on that stream calls `StreamReceiveComplete` with no receive outstanding.

With msquic 2.4+ multiple-receive completion accounting, a spurious zero-byte completion can permanently stop the stream from indicating receives: the stream stays open and healthy-looking, `setopt(active, N)` keeps returning ok, and received data is never delivered again.

### How it is hit in `{active, N}` usage

Two routine ways a receive lands in passive mode with nobody in `recv()`:

1. The disable at `{active, N}` count exhaustion is asynchronous (`StreamReceiveSetEnabled(FALSE)` from the callback), so a receive can slip in after the passive edge.
2. The stream's first data can arrive before the owner's initial `{active, N}` setopt.

Either arms `is_recv_pending`; the next re-arm correctly re-indicates the data, but from then on every `{active, N}` exhaustion/re-arm cycle fires a spurious completion. In our capture the stream died on the second spurious `StreamReceiveComplete(0)`.

### Evidence

Instrumented capture from a CI system running a credit ping-pong over one stream (fprintf tracing in the NIF receive path):

```
RECV   stream=X mode=passive len=548  pending=0   <- first data beats the initial {active,N}
LATE-RECV-PENDING stream=X len=548               <- QUIC_STATUS_PENDING, flag armed
REARM  stream=X pending=1                        <- initial {active,20}
REARM-COMPLETE0 stream=X                         <- correct: data re-indicated, delivered
RECV   stream=X mode=active ... pending=1        <- flag never cleared, all steady state
...
EDGE-DISABLE stream=X pending=1                  <- {active,N} exhaustion, flag STALE
REARM  stream=X pending=1
REARM-COMPLETE0 stream=X                         <- SPURIOUS #1 (stream survives)
... 18 more delivered messages ...
EDGE-DISABLE stream=X pending=1
REARM  stream=X pending=1
REARM-COMPLETE0 stream=X                         <- SPURIOUS #2
(no receive event ever indicated again; peer wedged with 45 messages queued)
```

Per-stream ledger: 1 `LATE-RECV-PENDING` vs 3 `REARM-COMPLETE0`.

Reproduction rate without the fix in our CI: ~1 in 15-25 runs of a site-restart test that cycles many `{active, N}` streams. With the fix, in the same test: `LATE-RECV-PENDING` count == `REARM-COMPLETE0` count exactly (175 == 175 in one full run, vs 154 vs 199 unfixed), and the outcome-level differential shows 0 stalls in 80 runs. The arming race itself is common (175 occurrences in a single passing run) - only the subsequent corruption was rare, which is why the stall was so intermittent.

### The fix

Clear `is_recv_pending` once the pending recv has been completed in the passive->active path of `set_stream_opt`. One line plus comment.

Related: fbcfe0a ("adapt to msquic 2.4 new recv mode") introduced this handling; the leak is in that path. 2cd5e59's handoff buffering is unaffected.

### DIFF
--- a/c_src/quicer_config.c
+++ b/c_src/quicer_config.c
@@ -1450,6 +1450,12 @@ set_stream_opt(ErlNifEnv *env,
               // Complete the pending recv (consuming 0 bytes) so the data is
               // re-indicated now that receives are re-enabled.
               MsQuic->StreamReceiveComplete(s_ctx->Stream, 0);
+              // The pending recv is completed; without this reset every later
+              // passive->active transition on the stream would issue a
+              // StreamReceiveComplete with no receive outstanding, which
+              // corrupts msquic's receive accounting and can permanently
+              // silence the stream.
+              s_ctx->is_recv_pending = FALSE;
             }
         }
       if (!set_owner_recv_mode(s_ctx->owner, env, optval))


qzhuyan:
sorry for the late reply. It took me some time to understand and reproduce the issue.

I managed to repoduce the issue with folllowing test code that the recv of `{quic, <<"ping_active_2">>, Stm, _} ` timeout but the fix in this PR does not seem to fix the issue. 
The problem is in the default behaviour, the dummy all of `StreamReceiveComplete ` will not reindicate the recv event,
I am now considering support 'multiple-receive' instead. 

``` erlang
tc_stream_passive_switch_to_active2(Config) ->
    Port = select_port(),
    Owner = self(),
    {SPid, Ref} = spawn_monitor(fun() -> echo_server(Owner, Config, Port) end),
    receive
        listener_ready -> 
            ok
    after 6000 ->
        ct:fail("timeout")
    end,

    {ok, Conn} = quicer:connect("localhost", Port, default_conn_opts(), 5000),
    {ok, Stm} = quicer:start_stream(Conn, [{active, false}]),
    {ok, 12} = quicer:send(Stm, <<"ping_passive">>),
    %% add delay to ensure pending.
    timer:sleep(500),
    {ok, <<"ping_passive">>} = quicer:recv(Stm, 0),
    {ok, 13} = quicer:send(Stm, <<"ping_passive2">>),
    %% add delay to ensure pending.
    timer:sleep(500),
    quicer:setopt(Stm, active, 1),
    quicer:setopt(Stm, active, false),
    %{ok, <<"ping_passive2">>} = quicer:recv(Stm, 0),
    quicer:setopt(Stm, active, 1),
    receive
        {quic, <<"ping_passive2">>, Stm, _} -> ok
    end,
    SPid ! done,
    ensure_server_exit_normal(Ref).
```

Some clarification below: 

> With msquic 2.4+ multiple-receive completion accounting, a spurious zero-byte completion can permanently stop the stream from indicating receives: the stream stays open and healthy-looking, setopt(active, N) keeps returning ok, and received data is never delivered again.

quicer is not yet using multiple-receive completion and it is using the default see [StreamReceiveComplete.md](https://github.com/microsoft/msquic/blob/main/docs/api/StreamReceiveComplete.md)


> How it is hit in {active, N} usage

> Two routine ways a receive lands in passive mode with nobody in recv(): 

>The disable at {active, N} count exhaustion is asynchronous (StreamReceiveSetEnabled(FALSE) from the callback), so a > receive can slip in after the passive edge.

receive will not slip in the callback as the stream is in the passive mode already. 

> set_stream_opt handles the passive->active transition by re-enabling receives and, when is_recv_pending is set, calling StreamReceiveComplete(0) so held data is re-indicated

This is no longer true in msquic 2.4 (default behaviour) if StreamReceiveComplete is already called.

jbevemyr:
Thanks for digging in, and for the corrections - you are right on the mode framing and I am happy to reword the commit message accordingly (quicer uses the default receive mode, not multiple-receive; the "multiple-receive completion accounting" sentence overstates).

I think we are looking at **two distinct holes in the same path**, and both of our observations are consistent:

**Your reproduction** arms `is_recv_pending` after a passive `recv/2` has already called `StreamReceiveComplete` for the indication. At the passive->active transition the flush relies on `StreamReceiveComplete(0)` re-indicating - which, as you say, msquic 2.4 default mode no longer does once the receive was already completed. The flush mechanism itself is broken there, and my one-liner cannot help: clearing the flag does not make the data come back. Agreed.

**Our production capture** is a different sequence, and the one-line fix demonstrably removes it. The arming we captured was not the exhaustion edge but the stream's *first* data arriving before the owner's initial `{active, N}` (our trace shows the pending flag set at stream start, mode still passive-initial, len=548). The re-arm then correctly delivered that data - but the flag stayed set forever, and every later `{active, N}` exhaustion/re-arm cycle issued a `StreamReceiveComplete(0)` with **no receive outstanding**. In default mode that spurious completion is not a no-op: our instrumented trace shows the stream going permanently silent right after the second spurious completion, which fits the completion being consumed against the *next* indication - and per your own point, a completed receive is never re-indicated, so the stream stays open and healthy-looking while delivering nothing.

Empirics for that second hole, from our CI (a site-restart test cycling many `{active, N}` streams over ~50 streams/run): unpatched, the stall reproduces at roughly 1 in 15-25 runs (two instrumented catches, per-stream ledger 1 pending-arm vs 3 completions, node-wide 154 vs 199); with this patch, **0 stalls in 80 runs** and the ledger balances exactly (175 == 175). So the invariant the patch restores - never call `StreamReceiveComplete` without a pending receive - has strong field evidence behind it, independent of the flush question.

**On the way forward**: supporting multiple-receive properly sounds like the comprehensive fix for both holes, and we would be glad to help validate it - our CI harness reproduces the production-shaped stall reliably enough for an 80-run differential, and we can run your branch through it when you have something ready. Until then, would you consider taking this one-liner as an interim hardening (with the commit message reworded to default-mode terms)? It closes the hole we hit in production; we are currently shipping it as a fork pin either way.

qzhuyan:
@jbevemyr thanks for quick reply. Yes I think we could merge and release.

qzhuyan (approved):

PR #441, opened 2026-08-18
Rework receive path onto msquic multiple-receive mode

jbevemyr:
This replaces the single-receive `is_recv_pending` /
`StreamReceiveSetEnabled(TRUE)` + `StreamReceiveComplete(0)` re-arm
machinery with msquic's multiple-receive mode, removing the class of
receive stalls behind the recent passive-recv fixes (including the one
merged in 8c0371e).

### The bug class this removes

In single-receive mode, when `recv/2` finds no buffered data it re-arms
with `StreamReceiveSetEnabled(TRUE)` and, if an indication was pending,
`StreamReceiveComplete(0)`. msquic does not re-indicate already-buffered
data after a zero-length completion in this state, so a request/response
peer can deadlock: the data sits in msquic's recv buffer, the app waits
for a `continue` signal that never comes. We hit this in production
(request/response over passive streams); 8c0371e fixed one instance, but
the underlying re-arm dance has more corners (passive->active
transitions had a related re-indication dependency).

### What this PR does

- Enables `StreamMultiReceiveEnabled` in every `QUIC_SETTINGS` built by
  `create_settings()` and as the library-wide default at `MsQuicOpen`,
  so every stream runs in multi-receive mode. There is deliberately no
  option to run the old mode: keeping both receive state machines alive
  is how this bug class survived. (Happy to discuss making it
  configurable if you see a need.)
- Replaces the pinned single-event state (`Buffers[2]`, `BufferCount`,
  `TotalBufferLength`, `is_recv_pending`) with a FIFO chain of receive
  segments in the stream ctx. Passive indications append to the chain
  and return `QUIC_STATUS_PENDING`; `recv/2` copies from the front and
  issues one cumulative `StreamReceiveComplete` for the bytes consumed.
  The re-arm dance is gone.
- Passive->active transitions deliver the pinned chain to the owner
  directly (multi-receive mode never re-indicates pending data),
  preserving the indication flags (FIN, 0-RTT) in the delivered
  message props.
- `recv/2` takes a stream handle ref while reading: chain buffers point
  into msquic-owned recv chunks, which are freed only by `StreamClose`
  (verified against msquic 2.5.x: reset/abort paths do not free chunks,
  and `QUIC_RECV_BUF_MODE_MULTIPLE` never retires chunks).
- Adds `is_shutdown_complete` to gate recv-path msquic calls after
  `SHUTDOWN_COMPLETE`. (`is_closed` cannot serve: it stays `TRUE` for
  the whole life of locally initiated streams.)

### Behavior notes

- Active mode is unchanged (copy + implicit full completion via
  `QUIC_STATUS_SUCCESS`), but in multi-receive mode indications arrive
  per data arrival rather than per drain cycle, so `{active, N}`
  consumes N in smaller steps under load. The `passive` message
  semantics are unchanged.
- `quicer:recv/2` semantics are unchanged (returns up to Count bytes,
  `{ok, not_ready}` + `continue` signal when empty).
- Flow-control behavior is equivalent: uncompleted chain bytes occupy
  the receive window exactly as undelivered buffered data did before.

### Validation

- Full CT: quicer_SUITE 105/105, connection 82/82, listener 90/90,
  snb 34/34, reg 13/13; eunit 26/26; all 55 PropEr properties
  (including the stateful models).
- ASan+UBSan build: quicer_SUITE + snb green, plus a dedicated
  ~360-stream stress (flood + partial reads + abort/close/
  passive-active flips mid-chain): zero sanitizer findings.
- Valgrind memcheck (JIT-checked BEAM): zero definite/indirect leaks,
  no error contexts in the NIF (remaining reports are msquic
  sendmsg-cmsg and OpenSSL noise).
- Soaked at system level in our event-streaming stack (volga): full
  interop matrix vs a second QUIC implementation, plus a 51-run
  restart/failover differential against the previous quicer build.

Happy to split the settings-forcing into its own commit or add a
regression CT case for the re-arm deadlock if you'd like.

### DIFF
--- a/c_src/quicer_config.c
+++ b/c_src/quicer_config.c
@@ -17,6 +17,7 @@ limitations under the License.
 #include "quicer_config.h"
 #include "quicer_internal.h"
 #include "quicer_queue.h"
+#include "quicer_stream.h"
 #include "quicer_tls.h"
 #include <msquichelper.h>
 
@@ -1282,6 +1283,11 @@ create_settings(ErlNifEnv *env,
       Settings->IsSet.PeerBidiStreamCount = TRUE;
     }
 
+  // The NIF receive path (recv chain + cumulative StreamReceiveComplete)
+  // requires msquic multi-receive mode on every stream.
+  Settings->StreamMultiReceiveEnabled = TRUE;
+  Settings->IsSet.StreamMultiReceiveEnabled = TRUE;
+
   return true;
 }
 
@@ -1432,30 +1438,14 @@ set_stream_opt(ErlNifEnv *env,
       if (ACCEPTOR_RECV_MODE_PASSIVE == s_ctx->owner->active
           && !IS_SAME_TERM(ATOM_FALSE, optval))
         {
-          // Switching from passive to active: msquic receives may have been
-          // disabled (e.g. by a partial passive recv, or when an {active, N}
-          // count was exhausted).
-          // Re-enable them unconditionally.
-          //
-          // Order matters: enable BEFORE completing the pending recv. In
-          // single recv-buffer mode, StreamReceiveSetEnabled(TRUE) will NOT
-          // queue a recv flush while a read is still pending
-          // (ReadPendingLength > 0). By enabling first and then completing
-          // recv with 0, the completion itself drives the flush
-          // (QuicStreamReceiveComplete re-flushes once ReceiveEnabled is
-          // TRUE), so delivery does not depend on the enable-time flush guard.
-          MsQuic->StreamReceiveSetEnabled(s_ctx->Stream, TRUE);
-          if (s_ctx->is_recv_pending)
+          // Switching from passive to active: deliver data still pinned in
+          // the recv chain (multi-receive mode never re-indicates it), then
+          // re-enable receives, which may have been disabled when an
+          // {active, N} count was exhausted.
+          stream_deliver_recv_chain(env, s_ctx);
+          if (s_ctx->Stream && !s_ctx->is_shutdown_complete)
             {
-              // Complete the pending recv (consuming 0 bytes) so the data is
-              // re-indicated now that receives are re-enabled.
-              MsQuic->StreamReceiveComplete(s_ctx->Stream, 0);
-              // The pending recv is completed; without this reset every later
-              // passive->active transition on the stream would issue a
-              // StreamReceiveComplete with no receive outstanding, which
-              // corrupts msquic's receive accounting and can permanently
-              // silence the stream.
-              s_ctx->is_recv_pending = FALSE;
+              MsQuic->StreamReceiveSetEnabled(s_ctx->Stream, TRUE);
             }
         }
       if (!set_owner_recv_mode(s_ctx->owner, env, optval))
--- a/c_src/quicer_stream.c
+++ b/c_src/quicer_stream.c
@@ -57,8 +57,6 @@ static QUIC_STATUS
 handle_stream_event_send_shutdown_complete(QuicerStreamCTX *s_ctx,
                                            QUIC_STREAM_EVENT *Event);
 
-static void reset_stream_recv(QuicerStreamCTX *s_ctx);
-
 static int
 signal_or_buffer(QuicerStreamCTX *s_ctx, ErlNifPid *owner, ERL_NIF_TERM sig);
 
@@ -740,13 +738,15 @@ recv2(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[])
     }
 
   TP_NIF_3(start, (uintptr_t)s_ctx->Stream, size_req);
-  enif_mutex_lock(s_ctx->lock);
 
-  if (!s_ctx->Stream)
+  // Hold a handle ref: chain buffers point into msquic recv chunks that
+  // are freed by StreamClose.
+  if (!LOCAL_REFCNT(get_stream_handle(s_ctx)))
     {
-      res = ERROR_TUPLE_2(ATOM_CLOSED);
-      goto Exit;
+      return ERROR_TUPLE_2(ATOM_CLOSED);
     }
+
+  enif_mutex_lock(s_ctx->lock);
 
   if (ACCEPTOR_RECV_MODE_PASSIVE != s_ctx->owner->active)
     {
@@ -754,31 +754,28 @@ recv2(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[])
       goto Exit;
     }
 
-  // We have checked that the Stream is not closed/closing
-  // it is safe to use the s_ctx->Stream in following MsQuic API calls
-
-  if (s_ctx->is_recv_pending && s_ctx->TotalBufferLength > 0)
+  if (s_ctx->recv_avail > 0)
     {
-      //
-      // Buffer is ready
-      //
       uint64_t size_consumed = recvbuffer_flush(
           s_ctx,
           &bin,
-          size_req > s_ctx->TotalBufferLength ? s_ctx->TotalBufferLength
-                                              : size_req);
+          size_req > s_ctx->recv_avail ? s_ctx->recv_avail : size_req);
       TP_NIF_3(consume, (uintptr_t)s_ctx->Stream, size_consumed);
-      reset_stream_recv(s_ctx);
 
-      if (size_consumed > 0)
+      s_ctx->is_wait_for_data = FALSE;
+
+      // Cumulative completion; in multi-receive mode msquic drains
+      // size_consumed bytes from the front of the pending data.
+      if (!s_ctx->is_shutdown_complete)
         {
-          s_ctx->is_wait_for_data = FALSE;
+          MsQuic->StreamReceiveComplete(s_ctx->Stream, size_consumed);
         }
 
-      // call only when is_recv_pending is TRUE
-      MsQuic->StreamReceiveComplete(s_ctx->Stream, size_consumed);
-
       res = SUCCESS(enif_make_binary(env, &bin));
+    }
+  else if (s_ctx->is_shutdown_complete)
+    {
+      res = ERROR_TUPLE_2(ATOM_CLOSED);
     }
   else
     { // want more data in buffer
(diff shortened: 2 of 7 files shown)


qzhuyan:
@jbevemyr , thanks again for the PR.

Do you still see problem in your CI or production with this commit merged: ? https://github.com/emqx/quic/commit/8c0371e8a05e6e686e569a83677c73b44f27d4b7

I would really like to have a test case cover the case that it intend to solve now even it is not 100% reproducible. 

Would be great if you could add a CT case for that or just describe it in terms of the quicer api calls for me to reproduce. 

FYI I have added some cases in the main branch (unreleased yet):

https://github.com/emqx/quic/commit/ed0b414e84b7048259d75620082b9337130ef270

The test case is still failing which is commented out: `tc_stream_passive_switch_to_active_with_recv_pending`, first look from the debug log is about the reordering in msquic that I havn't reported it.  Plan to have a deep look but I have to hook on to other high prio tasks and will come back later next week.

c_src/quicer_config.c
```diff
@@ -1282,6 +1283,11 @@ create_settings(ErlNifEnv *env,
       Settings->IsSet.PeerBidiStreamCount = TRUE;
     }
 
+  // The NIF receive path (recv chain + cumulative StreamReceiveComplete)
+  // requires msquic multi-receive mode on every stream.
+  Settings->StreamMultiReceiveEnabled = TRUE;
```
qzhuyan, line 1288: I agree to keep it for all settings.

c_src/quicer_stream.c
```diff
@@ -857,49 +843,82 @@ shutdown_stream3(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[])
...
+  s_ctx->recv_avail -= size;
+  return size;
+}
+
+// Copy the whole event payload into a fresh binary (active mode; the
+// event buffers are only valid during the callback).
+static void
+recvbuffer_copy_event(QUIC_STREAM_EVENT *Event, ErlNifBinary *bin)
+{
+  enif_alloc_binary(Event->RECEIVE.TotalBufferLength, bin);
```
qzhuyan, line 913: for safty check allocation fail

qzhuyan:
to my own notes:
1. utilise enif_realloc_binary instead of the new recvseg? 
2. check trace of passive->active msquic behaviour
3. What about the flags, shall we merge them in recvseg?
4. back pressure when in passive mode.

jbevemyr:
Hi William,

Status with 8c0371e: good news. With that fix in, we can no longer reproduce any stall that we can attribute to quicer's receive path. We spent this week root-causing the residual stalls in our CI and they turned out to be two problems outside quicer entirely:

1. A test-environment artifact: we run several processes on one host, each with its own msquic listener on the same port. msquic binds listeners to the wildcard address with SO_REUSEPORT, so all processes end up in one kernel steering group and flows get delivered to the wrong process, which answers with stateless resets the real owner cannot match. Nothing quicer can do about that; we carry a small msquic patch that binds the listener to its configured address in our test builds.

2. A bug in a different (non-quicer) QUIC client of ours that kept sending on a connection ID the server had retired via NEW_CONNECTION_ID retire_prior_to, and then could not match the resulting stateless resets.

So from our side there is no known outstanding issue against master with 8c0371e merged.

The recv-pending scenario, in quicer API terms:

1. Open a bidirectional stream in passive mode (`{active, false}`).
2. The peer sends a burst of data.
3. The owner calls `quicer:recv(Stream, N)` with N larger than what has arrived so far. The NIF stashes the partial data and completes the receive callback with `Complete(0)` to ask msquic for more, leaving the receive marked pending.
4. Before enough data arrives to satisfy N, the owner gives up on the blocking read and switches the stream to active mode with `quicer:setopt(Stream, active, true)`.
5. Stall: the stashed bytes are never delivered as active-mode messages, and depending on timing msquic does not re-indicate at all, because from its point of view the app still owns a pending receive that consumed zero bytes.

That is exactly the shape of your `tc_stream_passive_switch_to_active_with_recv_pending`, so I believe you are chasing the same hole rather than msquic reordering. The failure needs the race in step 3 to 4 (data in the stash, receive still pending, then the mode switch), which is why it is flaky as a CT case: whether the plain recv path or the pending path is hit depends on arrival timing.

How we reproduced it locally: a simple throughput harness where the sender writes continuously and the receiver alternates between `quicer:recv/2` with a size larger than typical delivery and a switch to active mode under load. On 611d473 that stalls within seconds with the pending receive holding stashed data (we had debug prints on that path, tagged LATE-RECV-PENDING, showing the receive stuck pending while the stash was non-empty).

The multi-receive rework in this PR avoids the problem structurally: the NIF never completes with 0 while holding stashed data, so there is no state where msquic believes the app owns a receive that the app has abandoned. That is also why the equivalent test passes on this branch.

Happy to turn the sequence above into a CT case against master if that helps; the tricky part is making step 3 to 4 deterministic, which probably needs a peer that sends a controlled partial chunk, then pauses until the setopt has happened, then resumes.

jbevemyr:
Follow-up on the CT case: I dug into `tc_stream_passive_switch_to_active_with_recv_pending` and found the root cause. It is not msquic reordering, and one correction to my earlier comment: this is not fixed by this PR either. Both master (ed0b414 + 8c0371e) and this branch fail the same way, so it is independent of the receive-path rework.

On my machine (Linux, msquic 2.5.10) your test is not even flaky: it fails 30/30. Bisecting it step by step isolated the trigger:

**The zero-length send is the poison.** `{ok, 0} = quicer:send(Stm, <<"">>)` permanently wedges the stream's send path, but only on a stream that has gone through the pend-then-switch sequence (passive stream, receive callback left pending, then `setopt(Stm, active, true)`). Everything sent after the zero-length send is accepted by the API but never reaches the peer. The peer therefore goes silent, and the client connection eventually dies of `connection_idle`.

Evidence chain, each variant run 5 to 10 times, all deterministic:

- Your test minus the empty send and the `recv(Stm, 0)` einval probe: passes. The pend-then-switch delivery itself is fine.
- Your test minus only the einval probe (empty send kept): fails. The einval probe is innocent.
- Empty send on a stream that never had a pended receive: passes. The zero-length send alone is harmless.
- 1-byte send instead of the empty send, same position: passes. It is specific to zero length.
- Empty send issued after the pended data has already been re-indicated and delivered: still fails. The stream stays in the poisoned state indefinitely; any later zero-length send kills it.
- Server-side tracing shows the server never receives the post-empty-send bytes ("ping_active"), so it is the sender's stream that wedges, not the receiver.

Why it looks flaky in CI: the failure needs the receive to actually be pending at `setopt` time. With `timer:sleep(100)` that depends on whether the echo has arrived, so on loaded CI machines the precondition is only sometimes met.

Deterministic CT case (fails 100% on master and on this branch, passes if you drop the zero-length send):

```erlang
%% @doc After a pended passive receive is switched to active mode, a
%% zero-length send must not wedge the stream send path. Deterministic
%% root cause of the flaky tc_stream_passive_switch_to_active_with_recv_pending.
tc_stream_pend_switch_then_empty_send(Config) ->
    Port = select_port(),
    Owner = self(),
    {SPid, Ref} = spawn_monitor(fun() -> echo_server(Owner, Config, Port) end),
    receive
        listener_ready -> ok
    after 6000 -> ct:fail("timeout")
    end,
    {ok, Conn} = quicer:connect("localhost", Port, default_conn_opts(), 5000),
    {ok, Stm} = quicer:start_stream(Conn, [{active, false}]),
    {ok, 15} = quicer:send(Stm, <<"ping_passiveeee">>),
    %% let the echo arrive and pend in the passive recv callback
    timer:sleep(100),
    ok = quicer:setopt(Stm, active, true),
    %% the zero-length send wedges the send path from here on
    {ok, 0} = quicer:send(Stm, <<"">>),
    %% the pended data is still re-indicated and delivered fine ...
    receive
        {quic, <<"ping_passiveeee">>, Stm, _} -> ok
    after 1000 -> ct:fail("pended data not delivered")
    end,
    %% ... but nothing sent after the zero-length send reaches the peer
    {ok, 11} = quicer:send(Stm, <<"ping_active">>),
    receive
        {quic, <<"ping_active">>, Stm, _} -> ok
    after 1000 -> ct:fail("send path wedged after zero-length send")
    end,
    SPid ! done,
    ensure_server_exit_normal(Ref).
```

Mechanism guess, to be verified against msquic: the pend is completed with `StreamReceiveComplete(0)` during the mode switch, and a later zero-length `StreamSend` request (queued with no payload, so `QueuedSendOffset` does not advance) seems to leave the stream marked app-blocked in `QuicStreamSendWrite` (`Right == QueuedSendOffset` -> `QUIC_FLOW_BLOCKED_APP`) in a state it never recovers from; without the earlier `StreamReceiveComplete(0)` the same zero-length send is harmless. I have not confirmed the msquic-internal part, but given that two quite different NIF receive implementations (master and this branch) reproduce it identically, the state machine issue looks like it lives on the msquic side of the boundary, possibly worth a minimal msquic-only repro and an upstream issue.

Practical options until then: reject or no-op zero-length sends in quicer:send (nothing legitimate is transmitted for them anyway unless FIN is set), which would also make the original test pass as written.

jbevemyr:
Closing the loop on the zero-length-send finding: I traced it all the way down, and it is an msquic send-scheduler bug, not anything in quicer's receive path on either branch.

Mechanism (confirmed with an instrumented msquic build): a zero-length `StreamSend` sets `QUIC_STREAM_SEND_FLAG_DATA` and queues a flush; the flush has nothing to write and never clears the flag; every later `StreamSend` then sees `DATA` already set and skips `QuicSendQueueFlushForStream`, so no flush is ever queued for the new data. If the connection is quiescent at that point, the stream is wedged until idle timeout. Concurrent send activity (ACKs, flow control updates, other streams) repairs the state incidentally, which is exactly why the test looked flaky: the pend-then-switch sequence only mattered because it shaped when the connection went quiescent.

Filed upstream with a standalone C repro (wedges without quicer, passes when the empty send is skipped): microsoft/msquic#6243.

Quicer-side mitigation in #442: treat zero-length sends without FIN or START as no-ops. With that guard, `tc_stream_passive_switch_to_active_with_recv_pending` passes 5/5 as written (it failed 30/30 on my machine without it), and the PR re-enables it plus adds a deterministic regression test. Full suite 110/110.

qzhuyan:
I think we could merge and I will polish it later. 

@jbevemyr  could you format the code with `make fmt`? 

you may see some warnings `include/quicer_types.hrl:31:10: syntax error before: '=>'` but should be fine.

jbevemyr:
Done — ran `make fmt` and pushed. Only c_src/quicer_stream.c changed (one call reflowed); the Erlang side was already clean, and I saw the harmless quicer_types.hrl warning you mentioned. Thanks for taking it in!

jbevemyr:
The mac-mesh (macos-14, 28, quictls) failure is `quicer_connection_SUITE:tc_conn_client_bad_cert` flaking with "server still running" (321/322 executed cases pass). Cert-path logic this PR does not touch, and the only delta since the green run is the whitespace-only fmt commit.

c_src/quicer_stream.c
```diff
@@ -1231,16 +1281,49 @@ get_stream_rid1(ErlNifEnv *env, int args, const ERL_NIF_TERM argv[])
...
+  uint32_t flags = 0;
+  for (QuicerRecvSeg *seg = s_ctx->recv_head; seg; seg = seg->next)
+    {
+      flags |= seg->flags;
+    }
+  uint64_t len = recvbuffer_flush(s_ctx, &bin, 0);
 
-  s_ctx->is_recv_pending = FALSE;
-  s_ctx->TotalBufferLength = 0;
+  ErlNifEnv *menv = enif_alloc_env();
```
qzhuyan, line 1310: a little bit heavy.

qzhuyan (approved):

PR #442, opened 2026-08-20
Treat zero-length sends without FIN or START as no-ops (avoids msquic send-path wedge)

jbevemyr:
A zero-length send handed to msquic can permanently wedge the stream send path: msquic sets `QUIC_STREAM_SEND_FLAG_DATA` for the empty request, the resulting flush has nothing to write and never clears the flag, and every later `StreamSend` skips queueing a flush because the flag is already set. Nothing sent afterwards reaches the peer and the connection eventually dies of idle timeout. Root cause with an instrumented trace and a standalone C repro is filed as microsoft/msquic#6243; full bisection story in #441 (comments).

This was the actual root cause of the flakiness in `tc_stream_passive_switch_to_active_with_recv_pending`: the empty send in that test wedges the stream whenever the connection is quiescent at that moment, which depends on timing. It was never a receive-path or reordering problem.

Changes:

- `quicer:send/3` and `quicer:async_send/3` treat a zero-length send without `FIN` or `START` as a no-op returning `{ok, 0}`. Nothing legitimate is transmitted for such a request anyway; empty sends carrying `FIN` (graceful shutdown) or `START` still go through. For async callers that pass `QUICER_SEND_FLAG_SYNC`, the `send_complete` message contract is preserved.
- Re-enables `tc_stream_passive_switch_to_active_with_recv_pending` (passes 5/5 with the guard, failed 30/30 without it on this machine).
- Adds `tc_stream_pend_switch_then_empty_send`, a deterministic regression test for the wedge.

Full `quicer_SUITE`: 110/110.

### DIFF
--- a/src/quicer.erl
+++ b/src/quicer.erl
@@ -777,19 +777,38 @@ send(Stream, Data) ->
     | {error, badarg | not_enough_mem | closed}
     | {error, stream_send_error, atom_reason()}.
 send(Stream, Data, Flag) when is_integer(Flag) ->
-    %% This is an sync send, set flag ?QUICER_SEND_FLAG_SYNC
-    case quicer_nif:send(Stream, Data, Flag bor ?QUICER_SEND_FLAG_SYNC) of
-        %% @todo make ref
-        {ok, _Len} = OK ->
-            receive
-                {quic, send_complete, Stream, false} ->
-                    OK;
-                {quic, send_complete, Stream, true} ->
-                    {error, cancelled}
-            end;
-        E ->
-            E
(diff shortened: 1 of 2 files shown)


qzhuyan:
@jbevemyr  thanks for the workaround and the testings!

do you agree this is just a workaround that may be reverted in the future when problem solved in msquic, as I think sending empty data should be legit, for example, use it for keep-alive while has no data exchange in application layer.

qzhuyan (approved):

qzhuyan:
merge but will revert this once get fixed in msquic

